PR #1, opened 2015-05-15
Nuspec fix

thefringeninja:
DLLs need to go in a lib\{target} folder


damianh:
Was _just_ doing same. You win :)

PR #7, opened 2016-02-18
add xunit.runner.visualstudio so tests can be run from visual studio

jruizaranguren:


damianh:
Hate the way this impacts the various csproj files... but I guess there is no other way.

PR #11, opened 2016-03-09
Remove Console.Writeline

thefringeninja:
should this be replaced with something else? Should it be logged at all?


danbarua:
LibLog pls

damianh:
It was boo-boo. LibLog will happen :)

PR #16, opened 2016-03-15
Correcting AssemblyName so MsSql tests are run

LeeCampbell:
Running `.\build.ps1` fails to find the `Cedar.EventStore.MsSql.Tests.dll` as it is not named as per the convention where the projectname is the folder and dll name.

This PR changes the AssemblyName of the MsSql test so that is matches the folder ad project name. This allows the `.\Build.ps1` to find the tests.

PR #17, opened 2016-03-15
Build script fail on Xunit test failures

LeeCampbell:
This PR is based off the PR #16 

* Wrap the call to xunit in an `exec` statement to throw when xunit exists with non-zero exit code
* Consistent formatting of  curlies and params 
* Commented out the un-built tests (`GetEventStore`, `Postgres` & `SqlLite`)
* Hide the dump of the `System.Xml.Linq` type information by capturing it as a variable.


damianh:
:+1:

PR #18, opened 2016-03-15
Fixes Race Condition with Subscription Poller

thefringeninja:


damianh:
Good stuff, cheers!

src/Cedar.EventStore/Infrastructure/AsyncLazy.cs
```diff
@@ -3,12 +3,13 @@
...
     using System;
+    using System.Threading;
     using System.Threading.Tasks;
 
     public class AsyncLazy<T> : Lazy<Task<T>>
     {
-        public AsyncLazy(Func<Task<T>> taskFactory, bool isThreadSafe) :
-            base(() => Task.Factory.StartNew(taskFactory).Unwrap(), isThreadSafe)
+        public AsyncLazy(Func<Task<T>> taskFactory)
+            : base(() => Task.Factory.StartNew(taskFactory).Unwrap(), LazyThreadSafetyMode.ExecutionAndPublication)
```
bartelink: `ExecutionAndPublication` is effectively the default mode for the `Lazy` ctor with one arg IIRC.

However my main issue with this snippet [and the article etc underpinning it] is the fact that a failure to initialize successfully results in the `AsyncLazy` 'caching' the failure - i.e. anyone what `await`s will see forever see the `Exception` rather than retrying. 

I have not looked at the surrounding code here but I'd venture you'll want to at least read about and potentially consider using the impl (not written by me even if it looks like it) in  http://stackoverflow.com/a/33946166/11635

PR #20, opened 2016-03-16
Insert events into MSSQL in deterministic order

### LINKED ISSUE #19: Cedar.EventStore.MsSql2008 doesn't store events in the correct order
What we're seeing is a mismatch between ordinal and streamversion for a stream.
Which leads to the highest ordinal number not having the highest streamversion.

ordinal

Full explantion and code: https://gist.github.com/bwaterschoot/77362620bbf058b58285

This happens because the events aren't stored in the correct order, link below:
https://social.msdn.microsoft.com/Forums/sqlserver/en-US/5cf485cc-a005-4f70-9cad-78b091d669cd/table-valued-parameters-order-of-records-to-stored-procedure?forum=transactsql

A solution would be in the ReadStreamForward to use StreamVersion instead of Ordinal as the orderby clause. Because you can't be sure SQL writes the event in the same order.

Probably would need to look at other Read sql statement to do something similar.

You can reproduce the issue using the code below, you might have to run it several times, because sometimes it does work ...

```c#    
    namespace Cedar.EventStore
    {
        using System;
        using System.Collections.Generic;
        using System.Threading.Tasks;
        using System.Linq;
        using Cedar.EventStore.Streams;
        using Shouldly;
        using Xunit;


        public partial class EventStoreAcceptanceTests
        {
        [Fact]
        public async Task Given_large_event_stream_can_be_read_back_in_pages()
        {
            using (var fixture = GetFixture("dbo"))
            {
                using (var eventStore = await fixture.GetEventStore())
                {
                    var eventsToWrite = CreateNewStreamEvents();

                    await eventStore.AppendToStream("stream-1", ExpectedVersion.NoStream, eventsToWrite);
                    

                    var readEvents = await new PagedEventStore(eventStore).GetAsync("stream-1");

                    readEvents.Count().ShouldBe(eventsToWrite.Length);
                }
            }
        }

        private NewStreamEvent[] CreateNewStreamEvents()
        {
            var eventsToWrite = new List<NewStreamEvent>();
            var largeStreamCount = 7500;
            for (int i = 0; i < largeStreamCount; i++)
            {
                var envelope = new NewStreamEvent(Guid.NewGuid(), $"event{i}", "{}", $"{i}");

                eventsToWrite.Add(envelope);
            }

            return eventsToWrite.ToArray();
        }
    }

    public class PagedEventStore
    {
        private readonly IEventStore _eventStore;

        public PagedEventStore(IEventStore eventStore)
        {
            _eventStore = eventStore;
        }

        public async Task<IEnumerable<StreamEvent>> GetAsync(string streamName)
        {
            var start = 0;
            const int BatchSize = 500;

            StreamEventsPage eventsPage;
            var events = new List<StreamEvent>();

            do
            {
                eventsPage = await _eventStore.ReadStreamForwards(streamName, start, BatchSize);

                if (eventsPage.Status == PageReadStatus.StreamDeleted)
                {
                    throw new Exception("Stream deleted");
                }

                if (eventsPage.Status == PageReadStatus.StreamNotFound)
                {
                    throw new Exception("Stream not found");
                }

                events.AddRange(
                    eventsPage.Events);

                start = eventsPage.NextStreamVersion;
            }
            while (!eventsPage.IsEndOfStream);

            return events;
        }

    }
}
```

LeeCampbell:
I believe this closes #19 

The change is based on a simple idea: `SELECT` queries do not have any guarantees of ordering unless you specify an `ORDER BY`. When any insert is being made, it is being done from a `dbo.NewStreamEvents` Table Value type parameter. This data is in an order, but that doesn't mean that it will be read in that order, which means it will not be inserted in that order. By simply specifying the `ORDER BY StreamVersion;` we can ensure that inserts will be in the order we expected.

I have added the test provided in #19, and was able to have it fail fairly consistently without the `ORDER BY` clauses. I have run the tests a dozen times since adding the `ORDER BY` clause and have not seen a failure.

Note this did _NOT_ fix the other two failing tests.

### DIFF
--- a/src/Cedar.EventStore.MsSql2008/SqlScripts/AppendStreamExpectedVersionAny.sql
+++ b/src/Cedar.EventStore.MsSql2008/SqlScripts/AppendStreamExpectedVersionAny.sql
@@ -20,7 +20,8 @@ BEGIN TRANSACTION AppendStream;
                         [Type],
                         JsonData,
                         JsonMetadata
-                   FROM @newEvents;
+                   FROM @newEvents
+               ORDER BY StreamVersion;
             END
        ELSE
            BEGIN
@@ -38,6 +39,7 @@ BEGIN TRANSACTION AppendStream;
                         [Type],
                         JsonData,
                         JsonMetadata
-                   FROM @newEvents;
+                   FROM @newEvents
+               ORDER BY StreamVersion;
            END
 COMMIT TRANSACTION AppendStream;
--- a/src/Cedar.EventStore.MsSql2008.Tests/Cedar.EventStore.MsSql2008.Tests.csproj
+++ b/src/Cedar.EventStore.MsSql2008.Tests/Cedar.EventStore.MsSql2008.Tests.csproj
@@ -88,6 +88,7 @@
     <Compile Include="..\SharedAssemblyInfo.cs">
       <Link>Properties\SharedAssemblyInfo.cs</Link>
     </Compile>
+    <Compile Include="EventStoreAcceptanceTests.ReadLongStreamInPages.cs" />
     <Compile Include="ExploratoryTests.cs" />
     <Compile Include="MsSqlEventStoreFixture.cs" />
     <Compile Include="MsSqlEventStoreTests.cs" />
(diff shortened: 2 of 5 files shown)


src/Cedar.EventStore.MsSql2008/SqlScripts/AppendStreamExpectedVersionAny.sql
```diff
@@ -20,7 +20,8 @@ BEGIN TRANSACTION AppendStream;
                         [Type],
                         JsonData,
                         JsonMetadata
-                   FROM @newEvents;
+                   FROM @newEvents
+               ORDER BY StreamVersion;
```
damianh: @yreynhout looks simple enough fix :)

damianh:
Lovely, thanks @LeeCampbell !

PR #21, opened 2016-03-16
[WIP] Edge case when reading past end of stream

thefringeninja:
Stream has one event in it. I `ReadStreamForwards(streamName, 0, 1, ct)`. Apply the first event to my aggregate. Then, I want to read the rest of the stream. `ReadStreamForwards(streamName, 1, int32.MaxValue, ct)`. I get sequence contains no elements. I feel that I should get an empty page here.

### DIFF
--- a/src/Cedar.EventStore/InMemory/InMemoryEventStore.cs
+++ b/src/Cedar.EventStore/InMemory/InMemoryEventStore.cs
@@ -358,7 +358,7 @@ namespace Cedar.EventStore
                 }
 
                 var lastStreamVersion = stream.Events.Last().StreamVersion;
-                var nextStreamVersion = events.Last().StreamVersion + 1;
+                var nextStreamVersion = (events.LastOrDefault()?.StreamVersion ?? lastStreamVersion) + 1;
                 var endOfStream = i == stream.Events.Count;
 
                 var page = new StreamEventsPage(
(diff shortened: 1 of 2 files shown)


thefringeninja:
What should the behavior be when reading backwards? i.e. what should `lastStreamVersion`, `nextStreamVersion`, and `endOfStream` be?

src/Cedar.EventStore/InMemory/InMemoryEventStore.cs
```diff
@@ -358,7 +358,7 @@ protected override void Dispose(bool disposing)
                 }
 
                 var lastStreamVersion = stream.Events.Last().StreamVersion;
-                var nextStreamVersion = events.Last().StreamVersion + 1;
+                var nextStreamVersion = (events.LastOrDefault()?.StreamVersion ?? lastStreamVersion) + 1;
```
damianh: Good stuff!

damianh:
> What should the behavior be when reading backwards? i.e. what should lastStreamVersion, nextStreamVersion, and endOfStream be?

What are the inputs `ReadStreamBackwards(streamName, ?, ?)` and what is in the eventstore at the time?

damianh:
I believe [this passing test](https://github.com/damianh/SqlStreamStore/blob/f04eef5c2a6dd51b07b9c5eea5e0b72c91919c9e/src/SqlStreamStore.AcceptanceTests/SqlStreamStoreAcceptanceTests.ReadStream.cs#L59) covers this scenario, so closing.

Let me know if not. 

Cheers.

PR #23, opened 2016-05-09
[WIP] Support for GES

thefringeninja:
Brings support for GES back into Cedar.

There are some outstanding issues. 

1. The acceptance tests make some assumptions about the underlying store:
  i. Nothing else is writing to the store while the tests run. GES uses streams internally to track stuff. So tests like `Given_empty_store_when_get_head_checkpoint_Then_should_be_minus_one` will naturally fail.
  ii. EventId idempotency. It is perfectly acceptable to send `ExpectedVersion: 5, EventId: 1` followed by `ExpectedVersion: 6, EventId: 1` to GES. This is because `EventId` is not like a PK in the traditional sense; instead, within a single stream, it is used along with the `ExpectedVersion` for the idempotency check.
 2. Backpressure still needs to be implemented. Unfortunately, GES does not use tasks when dispatching your events in a subscription. In order to ensure message ordering, an intermediate queue must be used. However, if you have a slow consumer, it is possible for this intermediate queue to fill up without bounds. The simplest solution ATM is to kill the subscription when the queue gets over a certain size. Open to suggestions here!

This PR will be rebased when master changes.


damianh:
Good stuff. Lets see if this can be maintained. Going to _release_ sql server and postgres support first.

damianh:
Closing this as I won't be supporting GES directly in this project. The rename has made it clear that this is for SQL implementations only. Please run this as a separate project if it scratches your itch.

PR #24, opened 2016-05-31
Add some locks so multiple writers are not failing all the time.

mihasic:
@damianh and the solution was so close


mihasic:
Might be good to test different kind of writes simultaneously, not just *VersionAny.

damianh:
:boom: that's awesome!

damianh:
Going to run some more load tests, thanks for the contrib.

damianh:
> Might be good to test different kinds of writes simultaneously, not just *VersionAny.

Yes I'll beef up the suite a lot.

PR #35, opened 2016-09-13
Support for delimited database identifiers

JefClaes:
We use a convention at work where the name of a database is something like 'Project-IntegrationTests'. 
Since the database identifiers are not enclosed in brackets  ([ ]), SQL Server throws a syntax error when initializing the store. I changed the test to use a delimited identifier instead of a regular one, and added brackes where necessary.


danbarua:
image

PR #40, opened 2016-09-29
Unify Initialization API Across Stores

thefringeninja:
I'm trying to avoid writing code that looks like this:

```
if (usingMsSql) {
	var store = new MsSqlStreamStore(...);

	startupActions.Add(ct => store.InitializeStore(false, ct))

	_store = _store;
} else if (usingAurora) {
	var store = new MySqlStreamStore(...);

	startupActions.Add(ct => store.InitializeStore(false, ct))

	_store = _store;
} else {
	_store = new InMemoryStreamStore();
}
```

Without this, encapsulation becomes extremely difficult.

### DIFF
--- a/src/SqlStreamStore/Infrastructure/StreamStoreBase.cs
+++ b/src/SqlStreamStore/Infrastructure/StreamStoreBase.cs
@@ -164,5 +164,9 @@ namespace SqlStreamStore.Infrastructure
             }
             base.Dispose(disposing);
         }
+
+        public abstract async Task InitializeStore(
+            bool ignoreErrors = false,
+            CancellationToken cancellationToken = default(CancellationToken));
     }
 }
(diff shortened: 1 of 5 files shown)


src/SqlStreamStore/Infrastructure/StreamStoreBase.cs
```diff
@@ -164,5 +164,9 @@ protected override void Dispose(bool disposing)
             }
             base.Dispose(disposing);
         }
+
+        public abstract async Task InitializeStore(
+            bool ignoreErrors = false,
+            CancellationToken cancellationToken = default(CancellationToken));
```
damianh: Virtual with no-op here. Means don't have to override in `InMemoryStreamStore`

PR #41, opened 2016-09-29
Unable to read writes from all subscription

thefringeninja:
I'm having a bug where I subscribe to all, write an event, then handle that event by writing to a command stream. The command handler on that stream attempts to read the stream that event came from and I get `PageReadStatus.StreamNotFound`.

However when I try to reproduce this with a failing test, it works for the regular stream and deadlocks on the all stream.

### DIFF
--- a/src/SqlStreamStore.AcceptanceTests/SqlStreamStoreAcceptanceTests.Subscriptions.cs
+++ b/src/SqlStreamStore.AcceptanceTests/SqlStreamStoreAcceptanceTests.Subscriptions.cs
@@ -686,6 +686,83 @@
             }
         }
 
+        [Fact]
+        public async Task Can_read_our_own_writes_on_all()
+        {
+            var source = new TaskCompletionSource<PageReadStatus>();
+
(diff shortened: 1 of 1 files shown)


src/SqlStreamStore.AcceptanceTests/SqlStreamStoreAcceptanceTests.Subscriptions.cs
```diff
@@ -686,6 +686,83 @@ public partial class StreamStoreAcceptanceTests
...
+                        source.SetResult(page.Status);
+                    };
+                    SubscriptionDropped subscriptionDropped = (reason, ex) =>
+                    {
+                        if(reason == SubscriptionDroppedReason.Disposed)
+                            return;
+                        source.SetException(ex);
+                    };
+                    using(var subscription = store.SubscribeToAll(
+                        Position.Start,
```
damianh: Needs to be null here. So a usability bug that can be fixed #42, so closing.

PR #42, opened 2016-09-30
SubscribeToAll refactor to make it clearer you are subscribing to the start or are continuing.

damianh:
Add new member `IReadonlyStreamStore.SubribeToAllFromStart(...)` and tweak `IReadonlyStreamStore.SubribeToAll(...)` such that it does not take a `long?` but a `long` to which you pass an known position to continue from.


src/SqlStreamStore.AcceptanceTests/SqlStreamStoreAcceptanceTests.Subscriptions.cs
```diff
@@ -107,8 +107,7 @@ public partial class StreamStoreAcceptanceTests
 
                     var receiveMessages = new TaskCompletionSource<StreamMessage>();
                     List<StreamMessage> receivedMessages = new List<StreamMessage>();
-                    using(store.SubscribeToAll(
-                        null,
+                    using(store.SubscribeToAllFromStart(
```
damianh: This replaces `SubscribeToAll(null, ....)` that is tripping people up.

src/SqlStreamStore/IReadonlyStreamStore.cs
```diff
@@ -143,7 +143,28 @@ public interface IReadonlyStreamStore : IDisposable
         ///     An <see cref="IStreamSubscription"/> that represents the subscription. Dispose to stop the subscription.
         /// </returns>
         IAllStreamSubscription SubscribeToAll(
-            long? fromPositionExclusive,
+            long continueAfterPosition,
```
damianh: Hope this makes it more obvious. One can still pass in a `Position.Start` but the parameter clearly states _continueAfter_ meaning you'll skip the first message at position 0.

damianh:
Not taking this approach, going to remain with long? and exclusive subscription.

PR #46, opened 2016-10-27
Subscriptions: Extends callback delegates to include the Subscription instance.

damianh:
Handling a `StreamMessageReceived` or a `SubscriptionDropped` callback will include a reference to the subscription instance.

### DIFF
--- a/src/SqlStreamStore/Subscriptions/AllStreamMessageReceived.cs
+++ b/src/SqlStreamStore/Subscriptions/AllStreamMessageReceived.cs
@@ -4,14 +4,13 @@
     using SqlStreamStore.Streams;
 
     /// <summary>
-    ///       Repesents a delegate that is invoked when a stream messages has been received in a subscription.
+    ///      Repesents a delegate that is invoked when a stream messages has been received in a subscription.
     /// </summary>
     /// <param name="subscription">
     ///      The source subscription.
     /// </param>
     /// <param name="streamMessage">
-    ///     The stream message.
-    /// </param>
+    ///     The stream message.</param>
     /// <returns>A task that represents the asynchronous handling of the stream message.</returns>
-    public delegate Task StreamMessageReceived(IStreamSubscription subscription, StreamMessage streamMessage);
+    public delegate Task AllStreamMessageReceived(IAllStreamSubscription subscription, StreamMessage streamMessage);
 }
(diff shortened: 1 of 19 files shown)


src/SqlStreamStore/Subscriptions/AllStreamMessageReceived.cs
```diff
@@ -0,0 +1,16 @@
...
+    /// <summary>
+    ///      Repesents a delegate that is invoked when a stream messages has been received in a subscription.
+    /// </summary>
+    /// <param name="subscription">
+    ///      The source subscription.
+    /// </param>
+    /// <param name="streamMessage">
+    ///     The stream message.</param>
+    /// <returns>A task that represents the asynchronous handling of the stream message.</returns>
+    public delegate Task AllStreamMessageReceived(IAllStreamSubscription subscription, StreamMessage streamMessage);
```
damianh: @yreynhout Had to add a separate delegate so the right subscription type is passed in the callback.

src/SqlStreamStore/Subscriptions/AllSubscriptionDropped.cs
```diff
@@ -0,0 +1,17 @@
...
+    /// </summary>
+    /// <param name="subscription">
+    ///     The source subscription.
+    /// </param>
+    /// <param name="reason">
+    ///     The subscription dropped reason.</param>
+    /// <param name="exception">
+    ///     The underlying exception that caused the subscription to drop, if one exists.
+    /// </param>
+    public delegate void AllSubscriptionDropped(IAllStreamSubscription subscription, SubscriptionDroppedReason reason, Exception exception = null);
```
damianh: @yreynhout Same for Subscription dropped callback.

yreynhout:
If one disposes a subscription inside the SubscriptionDropped callback, is the callback reentrant at that point? Meaning, we'll be notified again with a different reason (Dispose), correct?

damianh:
@yreynhout No. SubscriptionDropped callback will only ever be raised **once**. https://github.com/damianh/SqlStreamStore/blob/master/src/SqlStreamStore/Subscriptions/StreamSubscription.cs#L222-L225  So if disposed after it's been dropped for some other reason, it _won't_ be raised again 

One of the reasons is that I don't want any concurrent invocations here for different reasons at the same(ish) time.

damianh:
I've add some more info to the xml comments regarding that.

PR #49, opened 2016-11-07
Allow stream message json data to not be prefetched during stream reading.

damianh:
When reading a stream (or all) you may not be interested in the `JsonData` payload of every single message. As there is currently no limit on the size of the `JsonData` payload in the core (limitation is store specific), this PR allows the user to disable prefetching of that JsonData

The key changes are:
 - `IReadOnlyStreamStore.ReadAllForwards()`, `.ReadAllBackwards()`, `.ReadStreamForwards()`, `.ReadStreamBackwards()` all have an additional, optional parameter `prefetchJsonData = true`.
 - `StreamMessage.JsonData {get;}` property has been replaced with a method `Task<string> StreamMessage.GetJsonData()`. When prefetch is turned off this will result will result in a call to the server. 
 - `InMemoryStreamStore` simulates the async nature using a task offload if prefetching is turned off.
 - Prefetching is enabled by default.

Note there is a condition where the message _may_ be deleted between handling a `StreamMessage` instance and calling the `GetJsonData()`. The implementation here is to return a null when that occurs. 

This is a breaking change.


damianh:
Note: I'd like to separately support reads that allow server side of filtering by `StreamMessage.Type` or StreamId prefix.

src/SqlStreamStore/IReadonlyStreamStore.cs
```diff
@@ -18,14 +18,21 @@ public interface IReadonlyStreamStore : IDisposable
...
+        ///     The cancellation instruction.
+        /// </param>
         /// <returns>
         ///     An <see cref="ReadAllPage"/> presenting the result of the read. If all messages read have expired
         ///     then the message collection MAY be empty.
         /// </returns>
         Task<ReadAllPage> ReadAllForwards(
             long fromPositionInclusive,
             int maxCount,
+            bool prefetchData = true,
```
damianh: The additional optional parameter.

src/SqlStreamStore/InMemory/InMemoryStream.cs
```diff
@@ -13,8 +13,8 @@ internal class InMemoryStream
         private readonly GetUtcNow _getUtcNow;
         private readonly Action _onStreamAppended;
         private readonly Func<int> _getNextPosition;
-        private readonly List<InMemoryStreamMessage> _events = new List<InMemoryStreamMessage>();
-        private readonly Dictionary<Guid, InMemoryStreamMessage> _eventsById = new Dictionary<Guid, InMemoryStreamMessage>();
+        private readonly List<InMemoryStreamMessage> _messages = new List<InMemoryStreamMessage>();
```
damianh: Took the opportunity to clean up when I saw it.

damianh:
I should add, one of the other drivers of this was the HTTP API where we want to be able to GET streams with links to individual message json data (as opposed to embedding it).

PR #57, opened 2017-01-03
Improve the paging performance a bit

mihasic:
Skip MaxCount+1 message from being processed

### DIFF
--- a/src/SqlStreamStore.MsSql/MsSqlStreamStore.ReadAll.cs
+++ b/src/SqlStreamStore.MsSql/MsSqlStreamStore.ReadAll.cs
@@ -46,38 +46,45 @@ namespace SqlStreamStore
                             readNext);
                     }
 
-                    long lastOrdinal = 0;
                     while (await reader.ReadAsync(cancellationToken).NotOnCapturedContext())
                     {
-                        var streamId = reader.GetString(0);
-                        var streamVersion = reader.GetInt32(1);
-                        ordinal = reader.GetInt64(2);
-                        var eventId = reader.GetGuid(3);
-                        var created = reader.GetDateTime(4);
-                        var type = reader.GetString(5);
-                        var jsonMetadata = reader.GetString(6);
-
-                        Func<CancellationToken, Task<string>> getJsonData;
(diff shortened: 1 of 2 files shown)


damianh:
Thanks for the PR @mihasic . Can you comment in the `Files Changed` the specific area that was modified? I think indent change has mark too many lines as changed.

src/SqlStreamStore.MsSql/MsSqlStreamStore.ReadAll.cs
```diff
@@ -46,38 +46,45 @@ public partial class MsSqlStreamStore
...
-                        var streamVersion = reader.GetInt32(1);
-                        ordinal = reader.GetInt64(2);
-                        var eventId = reader.GetGuid(3);
-                        var created = reader.GetDateTime(4);
-                        var type = reader.GetString(5);
-                        var jsonMetadata = reader.GetString(6);
-
-                        Func<CancellationToken, Task<string>> getJsonData;
-                        if (prefetch)
+                        if(messages.Count == maxCount)
```
mihasic: Wrapped internals of loop in this if... else. Basically, last n+1 message is not fully read, but replaced with default(...)

src/SqlStreamStore.MsSql/MsSqlStreamStore.ReadStream.cs
```diff
@@ -115,34 +115,42 @@ public partial class MsSqlStreamStore
...
-                        var streamVersion1 = reader.GetInt32(0);
-                        var ordinal = reader.GetInt64(1);
-                        var eventId = reader.GetGuid(2);
-                        var created = reader.GetDateTime(3);
-                        var type = reader.GetString(4);
-                        var jsonMetadata = reader.GetString(5);
-
-                        Func<CancellationToken, Task<string>> getJsonData;
-                        if(prefetch)
+                        if(messages.Count == count)
```
mihasic: Same here

damianh: Ok, cheers!

PR #58, opened 2017-01-17
Append result verification

yreynhout:
Added placeholder tests that verify the result of the append behavior and the future NextExpectedVersion. The tests involving `ExpectedVersion.Any` are broken for MsSql (but not in memory) because those do not return the actual `CurrentVersion`, but `-1` instead. I favored making copies of the idempotency tests instead of mutating them (something you may want to rollback due to test explosion).

OT: Also fixed a few typo's in test names.


damianh:
Tests fixed in https://github.com/damianh/SqlStreamStore/commit/bc2b8bf1920ad06f39e61e6fd26884dce839c48f

PR #60, opened 2017-01-19
Ability to read an empty stream backwards

yreynhout:
... appears to be broken on the in memory implementation. This PR should fix that. Please review the values chosen to return upon reading an empty stream backwards (see tests).

### DIFF
--- a/src/SqlStreamStore/InMemory/InMemoryStreamStore.cs
+++ b/src/SqlStreamStore/InMemory/InMemoryStreamStore.cs
@@ -625,8 +625,12 @@ namespace SqlStreamStore
                     count--;
                 }
 
-                var lastStreamVersion = stream.Messages.Last().StreamVersion;
-                var nextStreamVersion = messages.Last().StreamVersion - 1;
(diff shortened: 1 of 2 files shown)


src/SqlStreamStore/InMemory/InMemoryStreamStore.cs
```diff
@@ -625,8 +625,12 @@ protected override Task<ReadStreamPage> ReadStreamForwardsInternal(string stream
...
                 }
 
-                var lastStreamVersion = stream.Messages.Last().StreamVersion;
-                var nextStreamVersion = messages.Last().StreamVersion - 1;
+                var lastStreamVersion = stream.Messages.Count > 0 
+                    ? stream.Messages[stream.Messages.Count - 1].StreamVersion
+                    : StreamVersion.End;
+                var nextStreamVersion = messages.Count > 0 
+                    ? messages[messages.Count - 1].StreamVersion - 1 
+                    : StreamVersion.End;
```
damianh: Looks fine @yreynhout

damianh:
Thanks for the PR lad.

PR #61, opened 2017-01-21
Gitbook docs

alexeyzimarev:
So @damianh decide yourself :) If you like the Gitbook, all you will need is to accept the PR and configure the build somewhere. I can explain how to build on Appveyor. Basically any build server would do, just run some commands from `docs.yml`


damianh:
I'm going to follow [squirrel.windows](https://github.com/Squirrel/Squirrel.Windows) and put the docs directly in the repo. Simpler all round I feel. Thanks for sharing the idea.

PR #63, opened 2017-02-07
InMemoryStreamStore.ReadStreamBackward() Bug

yreynhout:
ReadStreamBackward without prefetch returns each message twice as part of reading a page. This PR fixes that behavior and adds tests that verify correct behavior for both reading forward and backward as a regression countermeasure.


damianh:
@yreynhout Thanks for PR. Are you blocked by this?

yreynhout:
I have a note in my unit tests that says "Bug in SSS - Please put 1 once fixed SSS version is used". So, no, I can live with this behavior and without the fix for now.

PR #67, opened 2017-02-28
[WIP] Store mirroring

damianh:
A way to mirror a store to another store. It's just a projection really.

image


src/SqlStreamStore/Infrastructure/StreamStoreMirror.cs
```diff
@@ -0,0 +1,46 @@
+namespace SqlStreamStore.Infrastructure
+{
+    using System;
+    using System.Threading.Tasks;
+    using EnsureThat;
+    using SqlStreamStore.Streams;
+    using SqlStreamStore.Subscriptions;
+
+    public class StreamStoreMirror : IDisposable
```
damianh: This is the main place of interest. I think I may rename this to `StreamStoreCopy` as "mirror" implies _exact_ copy. In this approach, we're not interested in maintaining `position` parity. Also, the destination store may have independent writes applied to it.

src/SqlStreamStore.AcceptanceTests/SqlStreamStoreAcceptanceTests.Mirror.cs
```diff
@@ -0,0 +1,70 @@
...
+        {
+            using (var sourceFixture = GetFixture())
+            {
+                using (var desintationFixture = GetFixture())
+                {
+                    using (var sourceStore = await sourceFixture.GetStreamStore())
+                    {
+                        using (var destinationStore = await desintationFixture.GetStreamStore())
+                        {
+                            using (var mirror = new StreamStoreMirror(sourceStore, destinationStore))
```
danbarua: o_0

damianh:
Ye had to go looking that up. Lol

damianh:
I've decided to PR this as a Sample / Recipe. There are operational concerns (i.e. handling transient errors, being caught up) as well as other extensibility needs (metadata transformation, message enrichment etc) that I feel should be owned by the user and not somehow encapsulated in an opinionated design.

damianh:
Not going to taking this approach. Will revisit.

PR #69, opened 2017-03-06
Change signature of single message append to stream extension method

yreynhout:
... to return `Task<AppendResult>` as the multi message one does.


src/SqlStreamStore/StreamStoreExtensions.cs
```diff
@@ -182,7 +182,7 @@ public static class StreamStoreExtensions
         ///     The cancellation instruction.
         /// </param>
         /// <returns>A task representing the asynchronous operation.</returns>
-        public static Task AppendToStream(
+        public static Task<AppendResult> AppendToStream(
```
damianh: 👍

damianh:
Not a breaking change, but an backwards compatible API change. Going to bump version to 0.7.0

PR #70, opened 2017-03-06
Expose CurrentPosition as part of AppendResult

yreynhout:
... in both the inmemory and mssql implementation. Similar to what we've done for `CurrentVersion`.

There's a conceptual problem I don't really know how to solve, which is, what is the `CurrentPosition` to be when you append with `ExpectedVersion.NoStream` and **no messages** - we could return the `HEAD` position. As soon as there are events, both the `in memory` and the `sql` implementation seem to be doing _the right thing_.


src/SqlStreamStore/Infrastructure/StreamStoreBase.cs
```diff
@@ -31,6 +31,8 @@ public Task<AppendResult> AppendToStream(StreamId streamId, int expectedVersion,
                                    "{messageCount} messages.", streamId, expectedVersion, messages.Length);
             }
 
+            //TODO: What should I do here? I don't know the CurrentPosition at this point.
+            //Gut feeling says I should push this down to each implementation so they can do the right thing ...
```
yreynhout: @damianh This is what I want you to have a look at.

src/SqlStreamStore.MsSql/MsSqlScripts/CreateSchema.sql
```diff
@@ -13,6 +13,7 @@ BEGIN
         IdOriginal          NVARCHAR(1000)                          NOT NULL,
         IdInternal          INT                 IDENTITY(1,1)       NOT NULL,
         [Version]           INT                 DEFAULT(-1)         NOT NULL,
+		Position            BIGINT              DEFAULT(-1)			NOT NULL,
```
damianh: Note: breaking schema change. Need to consider migration...

src/SqlStreamStore.MsSql/MsSqlStreamStore.AppendStream.cs
```diff
@@ -15,6 +15,20 @@
...
+
+            public MsSqlAppendResult(int? maxCount, int currentVersion, long currentPosition)
+            {
+                MaxCount = maxCount;
+                CurrentVersion = currentVersion;
+                CurrentPosition = currentPosition;
+            }
+        }
+
         private static readonly Tuple<int?, int> NullAppendResult = new Tuple<int?, int>(null, -1);
```
yreynhout: I need to remove this.

src/SqlStreamStore/Streams/AppendResult.cs
```diff
@@ -3,10 +3,18 @@
     public class AppendResult
     {
         public readonly int CurrentVersion;
+        public readonly long CurrentPosition;
 
         public AppendResult(int currentVersion)
```
yreynhout: This ctor needs to go.

damianh:
> There's a conceptual problem I don't really know how to solve, which is, what is the CurrentPosition to be when you append with `ExpectedVersion.NoStream` and no messages - we could return the HEAD position. 

Yep return the HEAD position. I don't envisage any consumption problem with that. If it's `0` or `-1` or something else, I can envisage problems with consumers using that to read forward from.

damianh:
I'm still not sure what to do with breaking schema change.

yreynhout:
1. RE: database schema upgrade. Don't hold hands. Show the recipe. Let people decide when and where to execute it (more guidance/recipe rather than embedding in the api).
1. ETA for the ReadHead() is somewhere this week.

yreynhout:
Handled the edge case with a `ReadHeadPosition()`.

yreynhout:
@damianh Have you consider adding extended properties (effectively metadata) to your table (or even columns)? https://docs.microsoft.com/en-us/sql/relational-databases/system-stored-procedures/sp-addextendedproperty-transact-sql - this way, we could discover what schema we're dealing with.

The thing I envision is that if you don't call a new MsSqlStreamStore specific `Task DiscoverSchema(CancellationToken cancellationToken = default(CancellationToken))`, we assume the latest schema that came out of `Task CreateSchema(bool ignoreErrors = false, CancellationToken cancellationToken = default(CancellationToken))`. If we do call the `DiscoverSchemaVersion()` method as an explicit (usually one time) operation upon `startup`, we can use the extended property as a way to discover the schema we're dealing with (the absence of the property just means it's the schema we have today) and cache the result for the remainder of the `IStreamStore` singleton, or until a time we want to rediscover (perhaps in zero downtime or controlled upgrade scenario). The reason to keep the discovery to a minimum is obviously the performance impact. Having to call the `DiscoverSchemaVersion` method explicitly is as innate as calling `CreateSchema`, you have to be aware of what you're dealing with. Going forward the extended properties would become part of the schema (i.e. installed by a call to `CreateSchema`). It gives you the same level of fidelity as a separate version table gives you but leveraging a built-in sql server feature.

Whether to call it `DiscoverSchemaVersion` or `IsSchemaCompatible` is more about whether we can gracefully degrade capabilities or not. In this concrete example, if you don't upgrade the streams table, reading the streams forward and backward would have to do extra querying to compute the lastposition of a stream. Until you upgrade the schema, you're going to get a bit of degraded performance. Alternatively, we could just say that without the schema change the above code will not run on an older schema. You probably want to know that as early as possible, e.g. upon startup.

yreynhout:
Found a _potentially_ similar way of doing that in postgresql as well - https://www.postgresql.org/docs/current/static/functions-info.html#FUNCTIONS-INFO-COMMENT-TABLE

damianh:
Thanks for the suggest. I've made an attempt at an implementation over at #75 , if you could kindly review?

damianh:
#75 is merged. Please rebase a lgtm. 

I feel this is the last API related change before 1.0.

PR #73, opened 2017-03-25
Make PK non-clustered

mihasic:
Should support more non-blocking writers, since inserts would not fall on the same page all the time. Maybe, should make the index with Id a clustered one...


damianh:
Thanks Mike.

I am assuming this is non-breaking. However, I will need to implement a migration step.

mihasic:
It's not really done... I've created more as a reminder yet.

damianh:
Just for the observers, this will make a performance improvement.

damianh:
Did a bit of a manual load test - definite improvement.

PR #74, opened 2017-03-27
Use TemporarySqlLocalDbInstance as backing store for MsSqlStream store tests

bwaterschoot:
Using [TemporarySqlLocalDbInstance](https://github.com/martincostello/sqllocaldb/blob/master/src/SqlLocalDb/TemporarySqlLocalDbInstance.cs) instead of the [SqlLocalDbProvider](https://github.com/martincostello/sqllocaldb/blob/master/src/SqlLocalDb/SqlLocalDbProvider.cs) so when the tests finish running all files written to disk get cleaned up.

Went for the easiest option first, making it static and letting the finalizer handle cleaning up the unmanaged resources - as xUnit doesn't support the AssemblyCleanUp attribute or the same concept natively. 

Can change this to whatever your preference is ;-)

But at least when you run the tests there won't be any lingering files left.

### DIFF
--- a/src/SqlStreamStore.MsSql.Tests/MsSqlStreamStoreFixture.cs
+++ b/src/SqlStreamStore.MsSql.Tests/MsSqlStreamStoreFixture.cs
@@ -10,8 +10,9 @@ namespace SqlStreamStore
     {
         public readonly string ConnectionString;
         private readonly string _schema;
-        private readonly ISqlLocalDbInstance _localDbInstance;
         private readonly string _databaseName;
+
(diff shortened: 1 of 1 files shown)


bwaterschoot:
I guess it would be even better to make the instance name unique too, so you can have multiple test runner running the tests at the same time.

damianh:
> I guess it would be even better to make the instance name

Yeah the names would need to unique. I definitely use parallel test execution where ever possible (i.e. ncrunch).

damianh:
Merged, TY!

PR #75, opened 2017-04-05
MsSql Schema check

damianh:
This PR adds method a method to MsSqlStreamStore to check the schema version. It also is the start of v2 of schema in preparation for #70 .

### How to use:
```
var store = new MsSqlStreamStore(settings);
var result = await store.CheckSchema();

Console.Writeline(result.CurrentVersion);
Console.WriteLine(result.ExpectedVersion);
Console.WriteLine(result.IsMatch()); // true if CurrentVersion == ExpectedVersion
```

It is recommend that this is called by the developer during application startup. (Of course the developer will have to handle DB being unavailable at startup...). Failure to do so may result in runtime errors (or maybe data corruption) if a newer version of the library is used against an old schema (or old version used on a new schema).

### Implementation

An [extended property](https://technet.microsoft.com/en-us/library/ms190243(v=sql.105).aspx) is attached to the `{schema}.Streams` table with a name `version` and the schema version number (as a string). This is set when `CreateSchema()` is invoked. If the property is missing, this schema is considered to be `v1`. 

Thanks to @yreynhout for the implementation suggestion.


src/SqlStreamStore.MsSql.Tests/MsSqlStreamStoreAcceptanceTests.cs
```diff
@@ -86,5 +86,49 @@ private IDisposable CaptureLogs(ITestOutputHelper testOutputHelper)
...
+                }
+            }
+        }
+
+        [Fact]
+        public async Task When_schema_is_v1_then_should_not_match()
+        {
+            using (var fixture = new MsSqlStreamStoreFixture("dbo"))
+            {
+                using (var store = await fixture.GetStreamStore_v1Schema())
```
damianh: This method is actually internal. I'm going to keep all versions of schemas within the code base so I can test things like checks or compatibility.

src/SqlStreamStore.MsSql/MsSqlScripts/CreateSchema.sql
```diff
@@ -86,4 +86,13 @@ BEGIN
...
         JsonMetadata        NVARCHAR(max)                           NULL
     );
+END
+
+BEGIN
+    IF NOT EXISTS (SELECT NULL FROM SYS.EXTENDED_PROPERTIES WHERE [major_id] = OBJECT_ID('dbo.Streams') AND [name] = N'Version' AND [minor_id] = 0)
+    EXEC sys.sp_addextendedproperty   
+    @name = N'Version',
+    @value = N'2',
+    @level0type = N'SCHEMA', @level0name = 'dbo',
```
yreynhout: Should schema name (dbo) not be injected/substituted? I seem to recall one could specify the schema to operate on.

src/SqlStreamStore.MsSql/MsSqlStreamStore.cs
```diff
@@ -86,43 +91,91 @@ await command
...
-                            .NotOnCapturedContext();
-                    }
-                    else
+                    await command.ExecuteNonQueryAsync(cancellationToken)
+                        .NotOnCapturedContext();
+                }
+            }
+        }
+
+        internal async Task CreateSchema_v1_ForTests(CancellationToken cancellationToken = default(CancellationToken))
```
yreynhout: Minor suggestion - you could also link the V1 schema create script into the test project and make this a MsSqlStreamStoreTestFixture concern or even move that version of the schema to the tests entirely. I understand the desire to keep the script around, but don't necessarily see a need for them all to live in the MsSqlStreamStore project itself (MsSqlStreamStore tests seems close enough).

src/SqlStreamStore.MsSql/MsSqlScripts/CreateSchema.sql
yreynhout: Okay, I see where this is handled - ignore. https://github.com/damianh/SQLStreamStore/blob/4bb7a8669b2326e7e2a8ddc495c821369ac9c9b3/src/SqlStreamStore.Postgres/PgSqlScripts/Scripts.cs#L67

damianh:
Yeah "injection" is a "find & replace". It works, and the result is cached.
:)

damianh:
Reasonable, I'll see if I can move it out.



On 6 Apr 2017 1:19 a.m., "Yves Reynhout" <notifications@github.com> wrote:

*@yreynhout* commented on this pull request.
------------------------------

In src/SqlStreamStore.MsSql/MsSqlStreamStore.cs
<https://github.com/damianh/SQLStreamStore/pull/75#discussion_r110049370>:

> @@ -86,43 +91,91 @@ await command

                 using (var command = new
SqlCommand(_scripts.CreateSchema, connection))
                 {
-                    if(ignoreErrors)
-                    {
-                        await ExecuteAndIgnoreErrors(() =>
command.ExecuteNonQueryAsync(cancellationToken))
-                            .NotOnCapturedContext();
-                    }
-                    else
+                    await command.ExecuteNonQueryAsync(cancellationToken)
+                        .NotOnCapturedContext();
+                }
+            }
+        }
+
+        internal async Task
CreateSchema_v1_ForTests(CancellationToken cancellationToken =
default(CancellationToken))

Minor suggestion - you could also link the V1 schema create script into the
test project and make this a MsSqlStreamStoreTestFixture concern or even
move that version of the schema to the tests entirely. I understand the
desire to keep the script around, but don't necessarily see a need for them
all to live in the MsSqlStreamStore project itself (MsSqlStreamStore tests
seems close enough).

—
You are receiving this because you were assigned.
Reply to this email directly, view it on GitHub
<https://github.com/damianh/SQLStreamStore/pull/75#pullrequestreview-31193211>,
or mute the thread

yreynhout:
Minor typo in the description `var result = store.CheckSchema();` should be `var result = await store.CheckSchema();`

src/SqlStreamStore.MsSql/MsSqlScripts/CreateSchema.sql
damianh: Technically here https://github.com/damianh/SQLStreamStore/blob/4bb7a8669b2326e7e2a8ddc495c821369ac9c9b3/src/SqlStreamStore.MsSql/MsSqlScripts/Scripts.cs#L77 . Just a find / replace.

src/SqlStreamStore.MsSql/MsSqlStreamStore.cs
damianh: Will see if I can move it; maybe some infra coupling (i,e. the scripts "schema substitution").

damianh: So yeah there is a bit of infra layer around the script that I don't want to rewrite in the test project. I'm ok with keeping the script here for now, it's well hidden.

PR #64, opened 2017-02-09
[WIP] Update to support .Net Core

BennieCopeland:
Implementation work for issue #56.

I'm in the process of updating this to support .Net Core. Current effort is focused on targeting frameworks .Net 4.6 and .Net Standard 1.6.

Working on:
- [X] Restructure project into standard .Net Core structure (src and test folders)
- [X] Update SqlStreamStore project to support targeted frameworks
- [X] Update SqlStreamStore.Tests to support targeted frameworks and ensure tests are passing
- [x] Update SqlStreamStore.MsSql to support targeted frameworks
- [x] Update SqlStreamStore.MsSql.Tests to support targeted frameworks and ensure tests are passing
- [x] Update SqlStreamStore.Postgres to support targeted frameworks
- [x] Update SqlStreamStore.Postgres.Tests to support targeted frameworks ~~and ensure tests are passing~~
- [x] Update Example project to support targeted frameworks
- [x] Update cake build process

Not sure about:
- GhostDoc support
- ncrunch support
- ~~SharedAssemblyInfo.cs support~~
- ~~How rebasing will affect my current work~~

PR #77, opened 2017-04-12
Netstandard

damianh:
Continuing on from #64


damianh:
There seems to be intermittent failing tests that I can't reproduce locally :(

Going to merge and deal with CI issue after.

damianh:
@BennieCopeland Sincerest thanks for this contribution :)

PR #79, opened 2017-04-18
WIP: API Consistency work

BennieCopeland:
- Should `StreamMessage.GetJsonData()` repeat the database lookup, or cache the value on first call and return it on subsequent lookups?

### DIFF
--- a/src/LoadTests/ReadAll.cs
+++ b/src/LoadTests/ReadAll.cs
@@ -5,7 +5,7 @@
     using System.Threading;
     using System.Threading.Tasks;
     using EasyConsole;
-    using SqlStreamStore.Streams;
+    using SqlStreamStore;
 
     public class ReadAll : LoadTest
     {
--- a/src/SqlStreamStore.AcceptanceTests/SqlStreamStoreAcceptanceTests.AppendStream.cs
+++ b/src/SqlStreamStore.AcceptanceTests/SqlStreamStoreAcceptanceTests.AppendStream.cs
@@ -5,7 +5,7 @@
     using SqlStreamStore.Streams;
     using Xunit;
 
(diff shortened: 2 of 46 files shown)


damianh:
@BennieCopeland this is still WIP? It looks like it's all code cleanup (which is fine). I'd prefer to take the code clean and the API change in seperate PRs, so can focus on the change-at-hand.

BennieCopeland:
Still WIP. I'll go ahead and split the PRs and focus on the API changes.

damianh:
Flattening the API to the root namespace - what is driving that?

BennieCopeland:
Easier to use as a consumer when you only need to pull in a single namespace for basic functionality instead of also having to pull in Streams and Subscriptions. It's also easier to reason about what the public API is with it centrally located. Everything else is either optional items or implementation details located in the sub folders.

What I'm working on now is encapsulating the long's and int's for Position, StreamVersion, and ExpectedVersion into types that will provide a guided way to create the correct value. It will shield the user from knowing that -1 is `Position.End()` for example. It will only be used for inputs as I don't think it makes sense to have a `Position` type on `StreamMessage` and try to compare it against `Position.End()` which will always return false, even if it is the last message in the store. Also means the user doesn't have to figure out how to serialize the type since what they are getting back is a `long`/`int`.

damianh:
@BennieCopeland I would like this work to continue, however, I will have to target this for  version 2. It's a large breaking change from existing deployments (yes I know semver, but don't feel like breaking them now).

damianh:
@BennieCopeland I liked where this was going but to clean things up, am going to close this PR. If you get motivated to have another attempt, ping me in the slack channel.

PR #81, opened 2017-06-12
[WIP] - Building SqlStreamStore on linux

yreynhout:
I wanted the ability to build as much as possible of SqlStreamStore on linux, using dotnetcore. I've added warning messages to indicate what is and is not supported by the build script. Cake is still running on mono for the time being, but the rest is definitely dotnetcore based. The value is not tremendous at this point since neither nuget packaging (due to lack of .NET 4.6.1 SDK) nor running MsSql integration tests work on linux. Still, it's a better starting point for anybody wanting to contribute on linux.

### TODO

- [ ] Run tests against mssql using a docker container.


damianh:
Looking forward to this.

damianh:
You know what, perhaps delay this until NetStandard2 so I can drop the 4.6.1 specific build?

damianh:
Master now exclusively targets .netstandard 2.0. I believe cake still has a bit of dependency on nuget.exe (and thus .net desktop runtime) that is in process of being removed. I'm baking using Cake.CoreCLR on another project, will update here if it works out.

damianh:
@yreynhout I think this obsolete?

yreynhout:
Yes.

Issue #88, opened 2017-08-01
SqlStreamStore.MsSql 1.0.1 is broken


mihasic:
Default dependency of SqlStreamStore 1.0.1 does not exist, please publish 1.0.1 or a fixed SqlStreamStore.MsSql with reference to 1.0.0 of SqlStreamStore.

damianh:
https://www.nuget.org/packages/SqlStreamStore.MsSql/

![image](https://user-images.githubusercontent.com/57436/29017336-b645f238-7b56-11e7-86c0-9d596fb65df6.png)

Where do you see default dependency of of SqlStreamStore 1.0.1?

mihasic:
In assembly itself

damianh:
wtf

damianh:
crap 

![image](https://user-images.githubusercontent.com/57436/29017520-76a8b3a8-7b57-11e7-8697-6233e76c9b34.png)

damianh:
Will fix tomorrow. Unless you are using .net core in lambda, you don't need this patch update. Stick with 1.0.0

mihasic:
No rush for me at all

damianh:
Fixed in 1.0.2

PR #90, opened 2017-08-17
fix HasCaughtUp in subscriptions

mihasic:
- HasCaughtUp should be raised after push
- Prevent the raising twice fr…om single stream subscription.


mihasic:
@damianh , not sure if that's a valid change. Best for me would be to pass page.IsEnd to the [All]StreamMessageReceived, so I can commit the writes.
What's the idea?

damianh:
I think the changes are fine.

Not sure I want to expose page.IsEnd on `[All]StreamMessageReceived`, feels a bit leaky.

damianh:
I'll ship this package which should also fix #88

PR #91, opened 2017-08-18
migrate project to netstandard2.0

mihasic:
Migrated all the projects and build to netstandard 2.0

Had to remove support for sqllocaldb.

Maybe, it would be handy to extract sqllocaldb instances and tools to build script (using preinstalled sqllocaldb.exe utility). And database instance for tests can be feed using external settings file.

NOTE: still not really going to work in linux, due to SqlServer dependency and 'integrated security'.

to run load tests it's best to
```
cd src\LoadTest
dotnet publish -c Release -r win10-x64 -f netcoreapp2.0
.\bin\Release\netcoreapp2.0\win10-x64\LoadTests.exe
```

Anyway, please review.

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -1,5 +1,3 @@
-#tool "nuget:?package=xunit.runner.console&version=2.1.0"
-
 #addin "Cake.FileHelpers"
 
 var target          = Argument("target", "Default");
@@ -19,7 +17,6 @@ Task("RestorePackages")
     .Does(() =>
 {
 	DotNetCoreRestore(solution);
-	NuGetRestore(solution);
 });
 
 Task("Build")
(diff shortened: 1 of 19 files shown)


build.cake
```diff
@@ -43,13 +40,14 @@ Task("RunTests")
...
-        var settings = new ProcessSettings
+        var projectDir = "./src/" + testProject + "/";
+        var projectFile = testProject + ".csproj";
+        var settings = new DotNetCoreTestSettings
         {
-            Arguments = "xunit",
+            Configuration = configuration,
             WorkingDirectory = projectDir
         };
-        StartProcess("dotnet", settings);
```
damianh: `dotnet xunit` is different to `dotnet test`. Reason for the change?

mihasic: dotnet test feels more natural and does not require unstable package references. What's the reason for dotnet xunit?

damianh: It has richer output and more cli options. But yeah, it's not stable and had a problem with it on another project. This is fine for now.

damianh:
This actually works fine. Dual targeting is a PITA so will merge it sooner rather than later.

damianh:
Merged https://github.com/SQLStreamStore/SQLStreamStore/commit/eb5298cca01907f11c5846cbbc48a5ca69c00485 Thanks for PR @mihasic

PR #97, opened 2017-09-11
[WIP] MySql Adapter for Sql Stream Store

thefringeninja:
For Issue #29


src/SqlStreamStore.MySql/SqlStreamStore.MySql.csproj
```diff
@@ -0,0 +1,21 @@
...
+    <NoWarn>1701;1702;1705;1591</NoWarn>
+  </PropertyGroup>
+  <ItemGroup>
+    <EmbeddedResource Include="MySqlScripts\*.sql" />
+  </ItemGroup>
+  <ItemGroup>
+    <ProjectReference Include="..\SqlStreamStore\SqlStreamStore.csproj" />
+  </ItemGroup>
+  <ItemGroup>
+    <PackageReference Include="MySqlConnector" Version="0.26.2" />
```
bgrainger: :+1:

I'm the primary maintainer of MySqlConnector. Excited to see it be used more; please let me know if you encounter any issues.

src/SqlStreamStore.MySql/MySqlStreamStore.Delete.cs
```diff
@@ -0,0 +1,90 @@
...
+
+        private async Task DeleteStreamAnyVersion(
+            StreamIdInfo streamIdInfo,
+            CancellationToken cancellationToken)
+        {
+            using (var connection = _createConnection())
+            {
+                await connection.OpenAsync(cancellationToken);
+
+                using (var transaction = connection.BeginTransaction())
```
bgrainger: MySqlConnector extends the `MySqlConnection` API with a `BeginTransactionAsync(CancellationToken)` method: https://github.com/mysql-net/MySqlConnector/blob/b7a6cb2f9dc9c0db88703508f7393fd2efda684d/src/MySqlConnector/MySqlClient/MySqlConnection.cs#L27

There's also `CommitAsync` for line 56 below.

src/SqlStreamStore.MySql/MySqlStreamStore.ReadAll.cs
```diff
@@ -0,0 +1,206 @@
...
+            using (var connection = _createConnection())
+            {
+                await connection.OpenAsync(cancellationToken).NotOnCapturedContext();
+
+                var commandText = prefetch ? _scripts.ReadAllBackwardWithData : _scripts.ReadAllBackward;
+                using (var command = new MySqlCommand(commandText, connection))
+                {
+                    command.Parameters.AddWithValue("ordinal", ordinal);
+                    command.Parameters.AddWithValue("count", maxCount + 1); //Read extra row to see if at end or not
+                    var reader = await command
```
bgrainger: `MySqlDataReader` is `IDisposable`; best practice is to consume it in a `using` block so it's cleaned up deterministically.

src/SqlStreamStore.MySql/SqlStreamStore.MySql.csproj
damianh: Awesome @bgrainger ! Thanks for popping by and reaching out :)

damianh:
image

So close...

thefringeninja:
@bgrainger I would like to get your feedback on d3613f7. I was having problems running this as a single statement - the wrong stream would get deleted. I take it this is because mysql-net/MySqlConnector#382 is not implemented yet? If that is the case, I'm having a hard time understanding why the rest of this works 😕

bgrainger:
There are no problems with executing multiple commands in a batch that I am aware of.

I don't really understand the linked code, but is it possible you ended up executing something like this?

```
     DELETE Streams
       FROM Streams
      WHERE Streams.IdInternal = ?streamIdInternal;
...
     DELETE Streams
       FROM Streams
      WHERE Streams.IdInternal = ?streamIdInternal;
```

I.e., a single SQL statement that used the `?streamIdInternal` parameter twice? If so, you would need to change the code to use a unique parameter name each time.

If not, could you provide more details on what you were executing, what you observed, and what you expected to happen?

damianh:
Just closing to clear the PR Q. Will re-open on a new PR when actively worked on.

PR #99, opened 2017-09-13
Soften any assertions on Position such that it checks that it remains same or increases but do not assert specific value or sequentiality

damianh:
Position is generally implemented as an identity / sequence / auto-increment in the various stores. The value will always be increasing, but it is not guaranteed to be sequential. That is, there may be gaps.

Also, different stores will start at different positions values. i.e. MS SQL's identity starts at 0 whereas MySQL's auto increment starts at 1. It was discussed to normalize the behavior across all store implementation but was decided against - the position should be opaque to the end user and they should not derive any meaning from it's specific value. (Though they can derive meaning when comparing two values to see which is bigger or if they are equal.)

Each store fixture needs to define the `MinPositon` that is then used in tests. The tests have been updated to use this value as a base to see if the position has increased. Any assertions on a specific value of a position have been removed or altered.

### DIFF
--- a/src/SqlStreamStore.AcceptanceTests/SqlStreamStoreAcceptanceTestFixture.cs
+++ b/src/SqlStreamStore.AcceptanceTests/SqlStreamStoreAcceptanceTestFixture.cs
@@ -12,5 +12,7 @@ namespace SqlStreamStore
 
         public virtual void Dispose()
         {}
+
+        public abstract long MinPosition { get; }
     }
 }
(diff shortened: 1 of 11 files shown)


src/SqlStreamStore.AcceptanceTests/SqlStreamStoreAcceptanceTestFixture.cs
```diff
@@ -12,5 +12,7 @@ public abstract class StreamStoreAcceptanceTestFixture : IDisposable
 
         public virtual void Dispose()
         {}
+
+        public abstract long MinPosition { get; }
```
damianh: The new required property for store specific fixture.

damianh:
TC says it's good.

PR #101, opened 2017-09-20
Fix Test Reporting

thefringeninja:
Because the tests are all linked files, they have the same namespace / name, which means that Teamcity cannot differentiate them.

- use inheritance for acceptance tests so each gets a different name in test output
- dotnet xunit test runner
- updates to Cake 0.22 (allows c#6 syntax)


damianh:
Twerks in ncrunch

PR #105, opened 2017-10-13
[WIP] Build SqlStreamStore on Linux

pgermishuys:
Inspired by the work of Yves from the build-on-linux branch.

## Testing Locally on Linux

1. Pull the Microsoft SQL docker image

```
docker pull microsoft/mssql-server-linux
```

2. Start a Microsoft SQL docker instance

```
docker run -d -e 'ACCEPT_EULA=Y' -e 'SA_PASSWORD=SqlStreamSt0re' -p 1433:1433 microsoft/mssql-server-linux:latest
```

3. Run a build

```
./build.sh
```

## Dependencies

To build on linux the required dependencies are `Mono` and `Dotnet Core
2.0`.

## Additional work to be done

- Setup a CI environment for Linux

To add for example a [wercker](http://www.wercker.com/) build the following steps can be taken.

1. Add a `wercker` configuration file that will add a Microsoft SQL server as a service which will via docker links expose via environment variables the ip address as well as the port on which the Microsoft SQL service is available on.

The environment variables would be `MSSQL_SERVER_LINUX_PORT_1433_TCP_ADDR` and `MSSQL_SERVER_LINUX_PORT_1433_TCP_PORT`.

Contents: `wercker.yml`
```
build:
  services:
    - id: microsoft/mssql-server-linux
      env:
        ACCEPT_EULA: Y
        SA_PASSWORD: SqlStreamSt0re

  box:
    id: pgermishuys/sqlstreamstoretest:latest
  steps:
    - script:
        name: Build
        code: |
            ./build.sh 2>&1 | tee build-results.out
            ./inspect_test_results.sh
```

2. Add a script that will check the results of the tests.

Contents: `inspect_test_results.sh`
```
#!/usr/bin/env bash
test_result=$(grep FAIL build-results.out)
if [[ "$test_result" == "" ]] ; then
    exit 0
else
    exit 1
fi
``` 

3. Amend the `DockerInstance` to favour the environment variables over the defaults which are currently in place.

Contents: `DockerInstance`
```
private class DockerInstance : ILocalInstance
{
    private readonly string connectionString;
    public DockerInstance()
    {
        connectionString = BuildConnectionString();
    }
    private string BuildConnectionString(){
        var instanceAddress = Environment.GetEnvironmentVariable("MSSQL_SERVER_LINUX_PORT_1433_TCP_ADDR") ?? System.Net.IPAddress.Loopback.ToString();
[...]

PR #116, opened 2018-04-11
Postgres Implementation for SQL Stream Store

thefringeninja:
Implements #98

- Postgres driver for SqlStreamStore
- Builds correctly on linux (should cover #81 and #105)

Note: this requires docker to test correctly. The tests will pull and run the appropriate postgres version. It will not stop the container when it is finished. The container name is `sql-stream-store-tests-postgres`.


damianh:
Epic PR. Merging as is without squashing because dat commit count.

Am expecting subsequent PRs for fixes / perf / build / CI etc.

PR #119, opened 2018-04-28
Allow overriding the way a SqlConnection is created

CumpsD:
By default a connection is still created using `new SqlConnection(connectionstring)`, except the user now has the ability to override this behavior.


damianh:
Merged, thanks @CumpsD !

CumpsD:
@damianh it took a while before I managed to use this in our codebase, and I realised, it would've been better to let it make a `DbConnect` instead of a `SqlConnection`. I am wrapping a SqlConnection in a class (inheriting from DbConnection since SqlConnection is sealed)

What do you think? Does it make sense to change this to `DbConnection`?

Issue #121, opened 2018-05-02
It is possible for subscriptions to miss messages when chasing the tail under high concurrent load


spadger:
# Why

When inserting messages into SqlStreamStore, it is possible to get gaps in the stream, e.g. by a transaction rolling back, or by a transaction not yet ready for commit.

In the first case, the gap is legitimate, but in the second case, any subscribers to the messages should wait until  the gap has been filled.  It is not possible, upon seeing a gap to canonically determine the scenario under which a gap exists, so SqlStreamStore uses the following algorithm:

1. Read a page of messages
1. Check that every message's position is monotonically incrementing
1. If any message is missing, wait `DefaultReloadInterval`ms, and reread a batch
1. Return the new batch, without checking for gaps
The reason for ignoring gaps in step 4 is that if the gap is legitimate (e.g. a tx was rolled back), not to ignore gaps would cause the subscriber to continuously reload a page.

 
The underlying issue is that after the `DefaultReloadInterval` when a batch will be reloaded, subsequent messages may be written to the message queue.  These new messages may also contain gaps, but these new messages will never be checked for inconsistencies.

e.g., consider batch size is 100

First read.  5 Messages in the messages table, batch size is 100, and message 3 has not yet committed (but will be)

> `ReadAllForwards` sees messages [1, 2, *, 4, 5] // Message 3 is missing, so it will reload

Second Read. Now 10 messages from messages table, but message 7 has not been committed (but will be)

> `ReadAllForwards` sees messages [1, 2, 3, 4, 5, 6, *, 8, 9, 10] // Message 7 is missing

> `ReadAllForwards` will not re-validate this batch, and will pass it onto the subscription for processing, permanently hiding message 7 form the subscription.

# What
Alter the reloading strategy such that, upon reloading a page due to gaps, the reloaded page is re-validated, taking into account that a message may never exist:

* Load a page
* If a gap is detected:
  * Wait for a delay
  * Reload the page
  * Check for new gaps.  Any previously existing gaps can be ignored as abandoned messages (e.g. after a db tx rollback)
  * Loop until no new gaps exist

psibernetic:
In SQL Server at least you may be able to handle the uncommitted (2nd case) by using a sp_applock per sequence number as it is a reader/writer lock, you can block readers when it is being written, if you aren't blocked then it was a rollback. Unfortunately the other DBs may not have similar capabilities.

areicher:
[...]

PR #122, opened 2018-05-02
Safely deal with gaps in message positions. Fixes issue #121

spadger:
Fixes https://github.com/SQLStreamStore/SQLStreamStore/issues/121 (documentation is in there)


src/SqlStreamStore/Infrastructure/ReadonlyStreamStoreBase.cs
```diff
@@ -70,20 +71,55 @@ public abstract class ReadonlyStreamStoreBase : IReadonlyStreamStore
             // Check for gap between last page and this.
             if (page.Messages[0].Position != fromPositionInclusive)
             {
-                page = await ReloadAfterDelay(fromPositionInclusive, maxCount, prefetchJsonData, ReadNext, cancellationToken);
+                page = await ReloadAfterDelay(fromPositionInclusive, maxCount, prefetchJsonData, ReadNext, cancellationToken, new[] { fromPositionInclusive });
             }
 
-            // check for gap in messages collection
-            for(int i = 0; i < page.Messages.Length - 1; i++)
+            var missingMessagePositions = new HashSet<long>();
```
yreynhout: If I look at how gaps manifest themselves in MSSQL Server, i.e. skip by a 1000, it feels kind of wasteful to have an in memory representation of all these longs instead of e.g. `Range(from, to)`. Then again Range would probably be overkill if it'd be all individual positions. Still, how useful is it to log a 1000 numbers (further below)?Personally, I'm never going to be a fan of the _reload after delay_ way of handling this, but can't say I can come up with a better solution. The other thing that rubs the wrong way is that _if_ you pick a large enough `maxCount`, you're potentially going to be doing a lot reloads and also be blocking the caller. Whether that's to be expected / acceptable is another debate. The general recommendation is **not** to do a `maxCount = Int32.MaxValue`, so there's that. This is all more of a reflection than anything else - does not diminish the value of your PR.

src/SqlStreamStore/LongExtensions.cs
```diff
@@ -0,0 +1,21 @@
+﻿namespace SqlStreamStore
+{
+    using System;
+    using System.Collections.Generic;
+
+    public static class LongExtensions
```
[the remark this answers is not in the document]
yreynhout: Why not inline it? It's an implementation detail that shouldn't surface in the public API IMO.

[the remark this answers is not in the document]
spadger: Happy to - I have some unit tests around it to cover me from some off-by-one errors, but I guess they have done their job?

src/SqlStreamStore/Infrastructure/ReadonlyStreamStoreBase.cs
```diff
@@ -70,20 +71,55 @@ public abstract class ReadonlyStreamStoreBase : IReadonlyStreamStore
...
+
+                if (expectedNextPosition != actualNextPosition)
+                {
+                    var range = expectedNextPosition.RangeTo(actualNextPosition - 1);
+                    currentMissingPositions.UnionWith(range);
                 }
             }
 
-            return await FilterExpired(page, ReadNext, cancellationToken).NotOnCapturedContext();
+            currentMissingPositions.ExceptWith(previouslyMissingPositions);
```
yreynhout: Are we really sure that a gap _is real_ when it appears between _two reads_? Or are we guessing and making an assumption? Granted, it's highly unlikely but still possible.

src/SqlStreamStore/Infrastructure/ReadonlyStreamStoreBase.cs
yreynhout: We could have a lot of consumers of `ReadAllForwards()` - not overly worried about it but on the other hand measuring is knowing. As said, I feel very uncomfortable about the `ReloadAfterDelay()` - it makes assumptions about the consumer's behavior and willingness to accept this. Also an area where the MSSQL behavior starts to leak ...

src/SqlStreamStore/Infrastructure/ReadonlyStreamStoreBase.cs
spadger: Its probabilistic - the longer the gap, the surer we are, but a tx could technically last for an hour or longer before committing...(I say this to demonstrate a point - i'd question some life decisions if I ever saw this).
To be more certain would get very complicated, I'm guessing involving some two-phase message-committal 'stuff'

spadger: Bear in mind that this PR doesn't change any assumptions on the underlying code, just shores it up a bit, trading some perf for some correctness

src/SqlStreamStore/Infrastructure/ReadonlyStreamStoreBase.cs
spadger: I guess I could do one of a couple of things:
  * Leave the PR as-is (safe by default)
  * Add an opt-in to the new reload strategy(guaranteed safety within the retry window)
 
I don't want to add any retry limits etc, because you get the worst of both worlds - non-guaranteed message delivery combined with a slowdown

damianh:
I haven't forgotten about this issue. I think we need a bit more thinking around it. 💭

PR #126, opened 2018-05-07
Removing unspecified argument from formatted log.

tpresthus:
When debug logging is enabled, `StreamStoreBase.DeleteStream()` attempts to log the deletion with `streamId`, `expectedVersion` and `messageCount` as metadata. 

`messageCount` is however not available in the context of the method, thus resulting in an exception being thrown as the log statement can't be formatted.

This pull requests fixes it by removing `messageCount` from the log statement.


damianh:
Thanks for the PR @tpresthus  :)

PR #128, opened 2018-05-16
[WIP] MsSql - Self Managed Position

yreynhout:
This PR is an exploration into / an attempt at managing the `Position` ourselves instead of relying on the identity features of MSSQL Server. I fear there are serious downsides to doing this, i.e. a lot more unique index violations will occur.


damianh:
Going to close this as it has gone dormant. Also will need to work on the new MSSql schema if to be re-attempted. Cheers.

PR #129, opened 2018-05-20
Handle SetStreamMetadata idempotently

damianh:
In many cases a developer will lazily set the metadata of a stream. There are scenarios, especially when appending with StreamVersion.Any that they may set the same metadata over and over. This will increase the metadata stream length unnecessarily.

In another case, if users want to handling this idempotently, they have to invoke a read'n'check first. This has subtle race conditions.

This PR:
 - Adds a deterministic guid generator and tests
 - Add a `MetadataMessageIdGenerator` with a fixed GUID namespace.

### DIFF
--- a/src/SqlStreamStore.AcceptanceTests/SqlStreamStoreAcceptanceTests.StreamMetadata.cs
+++ b/src/SqlStreamStore.AcceptanceTests/SqlStreamStoreAcceptanceTests.StreamMetadata.cs
@@ -279,5 +279,26 @@
                 }
             }
         }
+
+
+        [Fact, Trait("Category", "StreamMetadata")]
+        public async Task When_set_metadata_with_same_data_then_should_handle_idempotently()
+        {
+            using (var fixture = GetFixture())
+            {
+                using (var store = await fixture.GetStreamStore())
+                {
+                    const string streamId = "stream-1";
+                    await store
+                        .SetStreamMetadata(streamId, maxCount: 2, maxAge: 30, metadataJson: "meta");
+                    await store
+                        .SetStreamMetadata(streamId, maxCount: 2, maxAge: 30, metadataJson: "meta");
+
+                    var metadata = await store.GetStreamMetadata(streamId);
+
+                    metadata.MetadataStreamVersion.ShouldBe(0);
+                }
+            }
+        }
     }
 }
--- a/src/SqlStreamStore.MsSql.Tests/SqlStreamStore.MsSql.Tests.csproj
+++ b/src/SqlStreamStore.MsSql.Tests/SqlStreamStore.MsSql.Tests.csproj
@@ -12,9 +12,6 @@
     <Compile Include="..\SqlStreamStore.AcceptanceTests\*.cs" />
   </ItemGroup>
   <ItemGroup>
-    <None Remove="SqlStreamStore.MsSql.Tests.v3.ncrunchproject" />
-  </ItemGroup>
-  <ItemGroup>
     <ProjectReference Include="..\SqlStreamStore.MsSql\SqlStreamStore.MsSql.csproj" />
   </ItemGroup>
   <ItemGroup>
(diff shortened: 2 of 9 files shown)


damianh:
I'm comfortable in declaring this as non-breaking.

src/SqlStreamStore.AcceptanceTests/SqlStreamStoreAcceptanceTests.StreamMetadata.cs
```diff
@@ -279,5 +279,26 @@ await store
...
+                {
+                    const string streamId = "stream-1";
+                    await store
+                        .SetStreamMetadata(streamId, maxCount: 2, maxAge: 30, metadataJson: "meta");
+                    await store
+                        .SetStreamMetadata(streamId, maxCount: 2, maxAge: 30, metadataJson: "meta");
+
+                    var metadata = await store.GetStreamMetadata(streamId);
+
+                    metadata.MetadataStreamVersion.ShouldBe(0);
```
damianh: The primary test. StreamVersion should not have been incremented and stay at 0.

src/SqlStreamStore.MsSql.Tests/SqlStreamStore.MsSql.Tests.csproj
```diff
@@ -6,13 +6,11 @@
     <GenerateRuntimeConfigurationFiles>true</GenerateRuntimeConfigurationFiles>
     <RootNamespace>SqlStreamStore</RootNamespace>
     <NoWarn>1701;1702;1705;1591</NoWarn>
+    <LangVersion>latest</LangVersion>
```
damianh: Slipped this in to all projects

src/SqlStreamStore.MsSql/MsSqlStreamStore.ReadStream.cs
```diff
@@ -87,7 +89,7 @@ public partial class MsSqlStreamStore
                 };
             }
 
-            using(var command = new SqlCommand(commandText, connection))
+            using (var command = new SqlCommand(commandText, connection, transaction))
```
damianh: Needed to pass this down because command wouldn't work if in a transaction scope. In other paths, this is null, which is fine.

damianh:
Would like to slip this into `v1.1.2`

src/SqlStreamStore.MsSql/MsSqlStreamStore.ReadStream.cs
yreynhout: I'm not completely following - are you saying that particular overload needs to be called even if transaction is null?

damianh: Because this transcation https://github.com/SQLStreamStore/SQLStreamStore/blob/4acea41e6ce4d338a0dfa76782bbee9cd4ada98e/src/SqlStreamStore.MsSql/MsSqlStreamStore.StreamMetadata.cs#L61 was getting associated with the connection, if I didn't pass it to this sql command I was getting sql exception.

```
System.InvalidOperationException: BeginExecuteReader requires the command to have a transaction when the connection assigned to the command is in a pending local transaction.  The Transaction property of the command has not been initialized.
   at System.Data.SqlClient.SqlCommand.ValidateCommand(Boolean async, String method)
   at System.Data.SqlClient.SqlCommand.RunExecuteReader(CommandBehavior cmdBehavior, RunBehavior runBehavior, Boolean returnStream, TaskCompletionSource`1 completion, Int32 timeout, Task& task, Boolean asyncWrite, String method)
   at System.Data.SqlClient.SqlCommand.BeginExecuteReader(CommandBehavior behavior, AsyncCallback callback, Object stateObject)
   at System.Threading.Tasks.TaskFactory`1.FromAsyncImpl[TArg1](Func`4 beginMethod, Func`2 endFunction, Action`1 endAction, TArg1 arg1, Object state, TaskCreationOptions creationOptions)
   at System.Threading.Tasks.TaskFactory`1.FromAsync[TArg1](Func`4 beginMethod, Func`2 endMethod, TArg1 arg1, Object state)
   at System.Data.SqlClient.SqlCommand.ExecuteReaderAsync(CommandBehavior behavior, CancellationToken cancellationToken)
```

PR #131, opened 2018-05-20
Allows setting of $deleted stream's metadata

### LINKED ISSUE #130: Purge $deleted
When a message or stream is deleted, a message is deleted to the $deleted stream. The primary purpose of this is to allow subscribers or (a future) mirroring capability to handle that. The current design retains these deleted messages indefinitely where as for most cases, one would only want to retain the for a period of time.

Throwing out some thoughts:

 1. Leverage the metadata infra to be able to set a `maxAge` and/or `maxLength`. Currently, the API does not allow direct manipulation of streams whose Id starts with `$` by design. I'd like to keep this in place so this approach would require an explicit API. The potential slight problem is that for `maxAge` the purge happens on read so a `ReadAllStreamForwards(null...)` would be required to actually remove the `$deleted` stream messages. If that never happens... `MaxCount` is handled on append, so that is not a concern.

 2. Expose a specific operation such as `PurgeDeleted(TimeSpan olderThan)` that the host application can invoke on it's own schedule. Should this have it's own implementation or underneath just reads the $deleted stream and let the `MaxAge` purge do it's thing?

damianh:
Allows settings the metadata (maxAge / maxCount) for `$deleted` stream via standard API.

Fixes #130 .

### DIFF
--- a/src/SqlStreamStore.AcceptanceTests/SqlStreamStoreAcceptanceTests.StreamMetadata.cs
+++ b/src/SqlStreamStore.AcceptanceTests/SqlStreamStoreAcceptanceTests.StreamMetadata.cs
@@ -300,5 +300,26 @@
                 }
             }
         }
+
+        [Fact]
+        public async Task Can_set_deleted_stream_metadata()
+        {
+            using (var fixture = GetFixture())
+            {
+                using (var store = await fixture.GetStreamStore())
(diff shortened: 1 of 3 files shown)


damianh:
This PR builds on #129 , hence the extra commits.

PR #132, opened 2018-05-20
Subscriptions: add CancellationToken to AllStreamMessageReceived delegate

damianh:
It's on the StreamMessageReceived delegate but was omitted from here from some reason.


damianh:
This is a breaking API change :(

PR #143, opened 2018-08-20
Improve performance of reading long streams

yreynhout:
After having fixed the append performance of long streams in the MsSql provider, it was time to assess the read performance of long streams. I noticed that reading forward started to slow down the further we were reading into the stream. Profiling revealed that an index scan was being performed that did an increasing number of reads the further we read. Tuning the SQL to use an index seek, and a fallback to index scan (in case messages are deleted from said stream during the reading operation), has improved the performance drastically (i.e. from around 30 minutes to around 5 minutes when reading around 2 million events in 1 stream as pages of a 1000 events each).

This PR changes the SQL scripts for reading streams of both the V2 and V3 of the schema - the database schema remains unchanged (except for the change below).

In addition, as per @thefringeninja 's suggestion, we've favoured using a clustered primary key (Position) on the Messages table and dropped the need for keeping track of an additional index on Position. This does change the V3 database schema!

### DIFF
--- a/src/SqlStreamStore.MsSql/ScriptsV2/ReadStreamBackward.sql
+++ b/src/SqlStreamStore.MsSql/ScriptsV2/ReadStreamBackward.sql
@@ -3,12 +3,24 @@
     DECLARE @streamIdInternal AS INT
     DECLARE @lastStreamVersion AS INT
 	DECLARE @lastStreamPosition AS BIGINT
+    DECLARE @position AS BIGINT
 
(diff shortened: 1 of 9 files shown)


global.json
```diff
@@ -0,0 +1,9 @@
+{
```
thefringeninja: This will almost certainly break it on my machine though :(

yreynhout: Going to remove it after I've rebased on top of @damianh 's work.

damianh: Yeah cake now donwloads and runs 2.1.400 if not already available on host.

PR #144, opened 2018-08-20
Build script and other maintenance things

damianh:
- Update to xunit 2.4.0
- Remove xunit dotnet cli tool (it's dead Jim).
- Simplify cake bootstrapper
- Some vscode dev time artifacts slipping in.
- Remove the parallel builds / tests / packaging for now so log messages are not interleaved.
- Update other dependencies.
- Platform / Test projects target `netcoreapp2.1`


damianh:
image

Looking good.

PR #145, opened 2018-08-21
Moved Common Test Package References to Own File

thefringeninja:
keeping all the common references e.g. xunit in sync among 4-5 projects is annoying.
needs to be tested with vs


yreynhout:
While I'm not opposed, this tends to change fairly slowly, so much that I wonder if it's worth the DRY-ness.

damianh:
This will mean manual managing of deps via text editing and loss of usage of tooling (`dotnet install` / nuget pacakge manager in VS). I don't think it's worth the dryness.

PR #148, opened 2018-08-21
Remove usage of SqlLocalDB and use sql server container in tests.

damianh:


damianh:
Works fine in ncrunch however seems slower than sqllocldb.

PR #149, opened 2018-08-22
[WIP] Reintroduce Parallel Testing

thefringeninja:
running tests in parallel via cake fails because of race conditions between the two mssql test projects when starting the container. This PR attempst to fix that.


damianh:
Lots of test failures with connection problems. I don't think this is going to work.

PR #154, opened 2018-08-26
fix build scripts

adamralph:


build.ps1
```diff
@@ -22,4 +17,4 @@ else{
 
 # Run Build
 
-dotnet run -p build\build.csproj -c Release --project $Target
\ No newline at end of file
+dotnet run -p build\build.csproj -c Release -- $args
```
adamralph: - `-p` already specifies the project, so currently `$Target` is being ignored
- all the args should be passed through, to allow `build.ps1 -h`, `build.ps1 -D`, etc.
- the target defaults to "default" if not specified

damianh: Lovely, thanks!

Issue #157, opened 2018-08-27
Remove Deleted Streams Table from Postgres Implementation


thefringeninja:
Would prefer not to have an extra table lying around

PR #158, opened 2018-08-27
Remove Deleted Streams Table from Postgres Impl

thefringeninja:
See #157


thefringeninja:
This causes `When_stream_has_max_count_and_append_exceeds_then_should_maintain_max_count` to fail :(

PR #163, opened 2018-09-02
switch targets to using enumerable inputs

adamralph:
It's a bit less code, and you get nicer output, with timings and success/failure messages for each project. E.g. for the `pack` target:

## Before

image

## After

image


damianh:
Lovely

PR #168, opened 2018-09-07
Run the compilation and tests inside docker.

damianh:
- Solution is compiled and test run in a container `microsoft/dotnet:2.1.401-sdk-alpine` 
- Database containers (SqlServer and Postgres), as before, are launched by the test fixtures as "sibling" containers.
- `build.cmd` that contains a single line for windows and `build.sh` that contains (basically) the same line for linux.
- Removed `build.ps1`
- dotnet test, ncrunch, resharper test runner still all work ootb
- could not get it to work via WSL because mounting the docker pipe / socket doesn't work and requiring tcp://localhost:2375 to be exposed was a config step too far.
- updated readme


damianh:
Build is green, whoo! https://travis-ci.org/SQLStreamStore/SQLStreamStore/builds/426095717

adamralph:
This is awesome. 😲

PR #170, opened 2018-09-10
Use minimal as the 'language'

thefringeninja:
The 'minimal' language contains everything we need to build

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,7 @@
-language: csharp
-mono: none
-dotnet: 2.1.401
+language: minimal
 sudo: required
 dist: trusty
 script: "./build.sh"
-before_install:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,12 +1,7 @@
-language: csharp
-mono: none
-dotnet: 2.1.401
+language: minimal
 sudo: required
 dist: trusty
 script: "./build.sh"
-before_install:
-- sudo service postgresql stop
```
damianh: We still need to stop the postgres service to prevent port clashing with the container instance, right?

thefringeninja:
Nope, because the minimal language doesn't have any database stuff as a
dependency.
https://docs.travis-ci.com/user/languages/minimal-and-generic/#minimal

PR #171, opened 2018-09-10
Reintroduce Parallel Testing

thefringeninja:


build/Program.cs
```diff
@@ -42,13 +44,16 @@ static void Main(string[] args)
...
+                () => Parallel.ForEach(new[]
+                    {
+                        "SqlStreamStore.Tests",
+                        "SqlStreamStore.MsSql.Tests",
+                        "SqlStreamStore.MsSql.V3.Tests",
+                        "SqlStreamStore.Postgres.Tests",
+                        "SqlStreamStore.Http.Tests"
+                    },
+                    project => Run("dotnet",
+                        $"test src/{project}/{project}.csproj -c Release -r ../../{ArtifactsDir} --no-build -l trx;LogFileName={project}.xml")));
```
adamralph: One thing to bear in mind is that you will lose the individual success/failure messages and timings from Bullseye for each test project with this change. Also, since you're invoking multiple `dotnet test` instances in parallel, you'll have 5 log outputs all interleaving with each other, so determining exactly what happened when could be a bit painful.

FWIW, I have parallel target execution [on the radar for Bullseye](https://github.com/adamralph/bullseye/issues/69). The output will probably also be interleaved in that case, but at least you'll still get a nice message of the form `Bulleseye/run-tests/SqlStreamStore.MsSql.Tests: Succeeded/Failed!` and a timing. In glorious colour of course. 🙂

adamralph: FYI Bullseye 2.0.0 RC3 is now available, with a new `--parallel` switch: https://www.nuget.org/packages/Bullseye/2.0.0-rc.3

I wanted to send a PR to upgrade to the RC, but I didn't want to do that before I was able to build locally: https://github.com/SQLStreamStore/SQLStreamStore/issues/184

thefringeninja:
This simply refuses to work on Travis. C'est la vie

thefringeninja:
May have another solution.

thefringeninja:
> System.Data.SqlClient.SqlException : A connection was successfully established with the server, but then an error occurred during the pre-login handshake. (provider: TCP Provider, error: 0 - No error information)

sigh

damianh:
Yeah that happens for me in ncrunch too. It's not going to happen. 

A final alternative would be to spin up a container for each test but that will cause other problems (SQL Server cointainers are not lightweight).

PR #172, opened 2018-09-10
Increase Travis Timeout

thefringeninja:
If travis doesn't receive any output for 10 minutes it fails the build. Some of our tests are taking almost that long: https://travis-ci.org/SQLStreamStore/SQLStreamStore/builds/426112365#L711

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: minimal
 sudo: required
 dist: trusty
-script: "./build.sh"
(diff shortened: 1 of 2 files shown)


thefringeninja:
You don't get any build output at all until the whole thing succeeds. 💢

damianh:
In addition to this, can we also output tests as they run?

https://github.com/dotnet/cli/issues/9229#issuecomment-388643284

PR #175, opened 2018-09-11
fix build arg passing

adamralph:
Relates to https://github.com/SQLStreamStore/SQLStreamStore/pull/168

Without this, `dotnet` was intercepting some options, e.g. `-h`.


damianh:
Good catch

PR #176, opened 2018-09-12
remove travis_wait

adamralph:
The build has been failing since #172 was merged due to the addition of `travis_wait`. This is because the build takes longer than 20 minutes, and `travis_wait` only extends the build by up to 20 minutes by default.

While a parameter can be passed to `travis_wait` to make the period longer, this shouldn't be required after adding `--verbosity=normal` to the `dotnet test` command, since each test result will be printed as it happens, and no individual test takes anywhere near 10 minutes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: minimal
 sudo: required
 dist: trusty
-script: travis_wait "./build.sh"
+script: ./build.sh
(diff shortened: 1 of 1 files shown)


adamralph:
Also, it's better not to use `travis_wait` if possible, to preserve the real time log output.

PR #177, opened 2018-09-12
Use intermediate Dockerfile

thefringeninja:
`-v %cd%:/repo` is causing problems on linux, since the container is running as root and creating files during the build, which then end up on the host's filesystem. Since these files have permissions 0600 you can't overwrite them without sudo - e.g. when you run a build via your IDE.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM microsoft/dotnet:2.1.401-sdk-alpine3.7 AS build
+FROM microsoft/dotnet:2.1.402-sdk-alpine3.7 AS build
 WORKDIR /src
 
 COPY ./src/*.sln ./
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,22 @@
+FROM microsoft/dotnet:2.1.401-sdk-alpine3.7 AS build
+WORKDIR /src
+
+COPY ./src/*.sln ./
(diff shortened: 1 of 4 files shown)


damianh:
I want to study this a bit before merging thx.

Dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM microsoft/dotnet:2.1.401-sdk-alpine3.7 AS build
```
damianh: Bump this to `4.0.2`.

Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+WORKDIR /src
+
+COPY ./src/*.sln ./
+COPY ./src/*/*.csproj ./
+RUN for file in $(ls *.csproj); do mkdir -p ./${file%.*}/ && mv $file ./${file%.*}/; done
```
damianh: Did you copy this from somewhere? Is there a description of why it is structured this particular way?

damianh: I presume there is some perf aspect / benefit of levergaging image layers, but a discription would be good. Thx

thefringeninja: When you do a copy operation in docker with a wild card, the file system hierarchy is not preserved - it's flattened instead: https://github.com/moby/moby/issues/15858

PR #178, opened 2018-09-12
Using 2.1.402-sdk-alpine image

damianh:


damianh:
Ignoring in favour of #177

PR #186, opened 2018-10-02
remove redundant dotnet restore commands

adamralph:
The restore for the `build` project is done as part of `dotnet run`, and the restore for the solution is done as part of `dotnet build` in the `build` target.

The restores in `Dockerfile` are redundant, and they add to the build time unnecessarily. The solution restore is especially redundant when running `build.cmd` in a way that does not run the `build` target. E.g. `build.cmd -n`, `build.cmd -D -I`, etc.

The removal of these lines may also make some of the other lines in the file redundant, but being a Docker noob I'm not 100% sure, so I'm leaving the rest as-is.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,15 +11,11 @@ RUN for file in $(ls *.csproj); do mkdir -p ./${file%.*}/ && mv $file ./${file%.
 
 COPY ./NuGet.Config ./
 
-RUN dotnet restore
-
 COPY ./src .
 
 WORKDIR /build
 
 COPY ./build/build.csproj .
-
-RUN dotnet restore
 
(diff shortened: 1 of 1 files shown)


thefringeninja:
It may increase build time on the server, but due to the way docker caching works it significantly reduces build times when working locally.

adamralph:
@thefringeninja ah, yeah, I see that now:

```
Step 7/13 : RUN dotnet restore
 ---> Using cache
 ---> a072cc83a833
```

This Docker stuff is cleverer than I realised. 😵

It's still redundant when the `build` target isn't being run, but only the first time `build.cmd` is invoked, so that's a small price to pay.

PR #189, opened 2018-10-04
Implement List Streams API

thefringeninja:
Implements #141.

This Pull Request adds a new API to `IReadonlyStreamStore`: `ListStream`. You can list streams starting with a string, ending with a string, or all streams, one page at a time. You may retrieve the next page by passing in the `continuationToken` in the current page. Currently, only paging forward is supported.


src/SqlStreamStore.MsSql/ScriptsV3/CreateSchema.sql
```diff
@@ -27,7 +28,16 @@ IF NOT EXISTS(
...
+
+IF NOT EXISTS(
+    SELECT *
+    FROM sys.indexes
+    WHERE name='IX_Streams_IdOriginal' AND object_id = OBJECT_ID('dbo.Streams', 'U'))
+BEGIN
+    CREATE NONCLUSTERED INDEX IX_Streams_IdOriginal
+        ON dbo.Streams (IdOriginal, IdOriginalReversed, IdInternal);
+END
+
```
damianh: I think this counts as a schema change?

src/SqlStreamStore.MsSql/ScriptsV3/Migration_v3_IdOriginalReversed.sql
```diff
@@ -0,0 +1,5 @@
+UPDATE [dbo].[Streams]
+SET [IdOriginalReversed] = REVERSE([IdOriginal]);
```
damianh: This might be a "big" db write. What if millions of rows?

src/SqlStreamStore.MsSql/ScriptsV3/CreateSchema.sql
```diff
@@ -11,6 +11,7 @@ BEGIN
     CREATE TABLE dbo.Streams(
         Id                  CHAR(42)                                NOT NULL,
         IdOriginal          NVARCHAR(1000)                          NOT NULL,
+        IdOriginalReversed  NVARCHAR(1000)                          NOT NULL,
```
damianh: Any concerns for the increase in size of the db?

src/SqlStreamStore.MsSql/ScriptsV3/CreateSchema.sql
thefringeninja: aha, so I need to add something for migration here?

src/SqlStreamStore.MsSql/ScriptsV3/Migration_v3_IdOriginalReversed.sql
thefringeninja: I can just move this into the migrator code. Will do

src/SqlStreamStore.MsSql/ScriptsV3/CreateSchema.sql
damianh: I believe so.

src/SqlStreamStore.MsSql/ScriptsV3/CreateSchema.sql
damianh: Computed columns are a sqlserver 2016 feature https://docs.microsoft.com/en-us/sql/relational-databases/tables/specify-computed-columns-in-a-table?view=sql-server-2017 . Dont think we should drop 2012R2 support just yet. (Am planning to do so with V2 to leverage more json things).

damianh: Otoh, wonder if we could support both at same time with feature detection...

src/SqlStreamStore.MsSql/ScriptsV3/Migration_v3.sql
```diff
@@ -14,6 +14,21 @@ IF NOT EXISTS (SELECT 1 FROM SYS.COLUMNS WHERE OBJECT_ID = OBJECT_ID(N'[dbo].[St
...
+    ADD [IdOriginalReversed] NVARCHAR(1000) NOT NULL DEFAULT '';
+
+IF NOT EXISTS(
+    SELECT *
+    FROM sys.indexes
+    WHERE name = 'IX_Streams_IdOriginal'
+      AND object_id = OBJECT_ID('dbo.Streams', 'U'))
+  CREATE NONCLUSTERED INDEX IX_Streams_IdOriginal
+    ON dbo.Streams (IdOriginal, IdOriginalReversed, IdInternal);
+
```
damianh: 👍

src/SqlStreamStore.MsSql/ScriptsV3/CreateSchema.sql
thefringeninja: We can, but I think this could make upgrading your sql server version a bit painful

PR #190, opened 2018-10-09
update Bullseye to 2.1.0

adamralph:


adamralph:
Something weird is happening on Travis CI. Judging by the colours in the output, the old version of Bullseye is being used. Is Docker doing some weird over-aggressive caching?

adamralph:
Ah, I know what the problem is. Bullseye doesn't know it is running in Travis CI, because the Travis CI environment variables are not present in the container. 😏

adamralph:
FYI in order for Bullseye 2.1.0+ to choose the appropriate palette, you need to propagate the `TRAVIS_OS_NAME` env var from the host to the container.

I'll leave it to you to decide whether you want to do that (either in this PR or later), since it's worth upgrading to 2.1.0 regardless, and not least because I'm a total Docker noob and don't know how to do it.

Also, I'm considering adding CLI options in a future version to force Bullseye to work in the mode for a specific CI system, i.e. `--appveyor`, `--travis-ci`, `--team-city`. That would make this easier, since then you just need to add the option here: https://github.com/SQLStreamStore/SQLStreamStore/blob/5f98e536be986436629413eb36986351fe1f95a5/.travis.yml#L4

thefringeninja:
What you do is pass the environment variable to docker as a build arg, then back to bullseye as you would on the shell. You can see that in action here: https://github.com/SQLStreamStore/SQLStreamStore.HAL/blob/master/Dockerfile

thefringeninja:
I went ahead and made a similar change here: https://github.com/SQLStreamStore/SQLStreamStore.HAL/pull/23/files
Not sure if it's working because the build output looks the same to me

adamralph:
@thefringeninja I left [some comments](https://github.com/SQLStreamStore/SQLStreamStore.HAL/pull/23#pullrequestreview-164126363) in the HAL PR.

I'm not sure it's worth the hassle here. I suggest merging this PR as-is, just to get the upgrade to 2.1.0.

In 2.2.0 we will have the [new `--travis` option](https://github.com/adamralph/bullseye/issues/127) so when we upgrade to that we can just add `--travis` to the `build.sh` call in `.travis.yml`.

PR #196, opened 2018-10-23
fix whitespace

adamralph:
I just had to fix that before continuing with my talk here at .NET Conf Switzerland 2018!


damianh:
Lol!

PR #197, opened 2018-10-23
Make streamId a sql parameter of type SqlDbType.Char

cheesebridge:
In our project we saw serious performance issues when having millions of streams.
This issue was caused by a CONVERT_IMPLICIT() function that was always executed when searching streams by streamId. 
Our streamId was a regular Guid and the parameter type got translated to a nvarchar(36)
The result was an index scan on the streams table instead of an index seek.

In this PR, all streamId parameters are typed as SqlDbType.Char with length 42 because the streamid column on the streams table is declared as CHAR(42).
Now an index seek can be done.

image

Before
image

After
image

### DIFF
--- a/src/SqlStreamStore.MsSql/MsSqlStreamStore.AppendStream.cs
+++ b/src/SqlStreamStore.MsSql/MsSqlStreamStore.AppendStream.cs
@@ -131,7 +131,7 @@
         {
             using(var command = new SqlCommand(_scripts.AppendStreamExpectedVersionAny, connection, transaction))
             {
-                command.Parameters.AddWithValue("streamId", sqlStreamId.Id);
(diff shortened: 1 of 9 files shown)


thefringeninja:
Just waiting on CI now... I had b0rked the build but just fixed it

damianh:
I'm going to cherry pick this across to a 1.1.3 release branch.

damianh:
v1.1.3 is now on nuget.org.

yreynhout:
Good stuff! Not sure we can improve on this but Index Scan is generally not a good thing - we should strive for Index Seeks (but as said, I don't know whether that is possible here).

cheesebridge:
You're welcome. It's a great product 👍 

@yreynhout Well this PR turns the index scan into an index seek as you can see in the 'after' screenshot of the query execution plan.  ;-)

yreynhout:
My bad 😂

PR #208, opened 2018-11-03
[wip] switch to MinVer

adamralph:
Sorry about the noise, I just wanted to test MinVer `1.0.0-alpha.12` on your Travis build. Hopefully this is the last alpha.

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -1,7 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project>
   <PropertyGroup>
-    <VersionPrefix>1.2.0</VersionPrefix>
     <Authors>Damian Hickey</Authors>
     <PackageProjectUrl>https://github.com/SqlStreamStore/SqlStreamStore</PackageProjectUrl>
     <PackageLicenseUrl>https://github.com/SqlStreamStore/SqlStreamStore/blob/master/LICENSE</PackageLicenseUrl>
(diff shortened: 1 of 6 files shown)


src/Directory.Build.props
```diff
@@ -1,13 +1,20 @@
...
     <PackageLicenseUrl>https://github.com/SqlStreamStore/SqlStreamStore/blob/master/LICENSE</PackageLicenseUrl>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <PackageTags>cqrs;event-sourcing;event-store;stream-store</PackageTags>
     <NoWarn>1701;1702;1705;1591</NoWarn>
     <LangVersion>latest</LangVersion>
+    <MinVerTagPrefix>v</MinVerTagPrefix>
+    <MinVerMinimumMajorMinor>1.2</MinVerMinimumMajorMinor>
+    <BUILD_NUMBER Condition="'$(BUILD_NUMBER)' == ''">$(TRAVIS_BUILD_NUMBER)</BUILD_NUMBER>
+    <BUILD_NUMBER Condition="'$(BUILD_NUMBER)' == ''">0</BUILD_NUMBER>
+    <MinVerBuildMetadata>build.$(BUILD_NUMBER)</MinVerBuildMetadata>
```
adamralph: FWIW, I thought it would communicate the portability of the build better if `BUILD_NUMBER` was the env var that controlled the build number rather than `TRAVIS_BUILD_NUMBER`, but still falling back to that if not set.

adamralph:
🎉 😭 

And to prove the build metadata works:





FYI that last image is actually taken from the NU5105 "SemVer 2.0.0" warning, but that can be easily suppressed for all projects after https://github.com/SQLStreamStore/SQLStreamStore/pull/207 is merged. FWIW, before beta, I'll ensure the full version with build metadata is shown in the output when `MinVerVerbose` is set to `true`.

PR #209, opened 2018-11-03
don't publish without running tests

adamralph:
Unless you don't mind publishing without running the tests first?

Assuming publishing should only happen after running the tests, it should not be the responsibility of the `default` target to set up this dependency via it's own dependencies, which also requires declaring those dependencies in the right order. Rather, the `publish` task should declare the dependency.

I guess you may want to manually run `publish` locally on it's own, having run the tests earlier, but in that case, you wouldn't want to rebuild either, so the correct way to run that is with `./build.cmd -s publish` or `./build.sh -s publish` (`-s` is shorthand for `--skip-dependencies`).

### DIFF
--- a/build/Program.cs
+++ b/build/Program.cs
@@ -56,7 +56,7 @@ namespace build
                     "SqlStreamStore.Http"),
                 project => Run("dotnet", $"pack src/{project}/{project}.csproj -c Release -o ../../{ArtifactsDir} --no-build --version-suffix {versionSuffix}"));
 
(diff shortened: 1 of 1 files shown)


adamralph:
Come think of it, this may not be what you want. After all, `publish` has is no _physical_ dependency on `run-tests`. I.e. `publish` does not require `run-tests` to have been run in order for it to succeed.

You may not _want_ to do that, but there are other ways of expressing that. I have some other ideas but I'll run them by you another time.

PR #212, opened 2018-11-06
remove clean target

### LINKED ISSUE #210: The clean target removes artifacts/.gitignore
Personally I would get rid of the `clean` target. It doesn't _really_ do a "clean", e.g. everything in `bin` and `obj` is left behind. A real clean is `git -fxd`.

Also, it's only a dependency of `default` anyway, so if you run just `./build.sh run-tests` or `./build.sh pack`, the artifacts are not cleaned first, they are just overwritten.

And on Travis CI, `clean` serves no function, because each build is on a clean VM anyway.

FWIW, I no longer have a `clean` target in any of my projects. If you'd like me to send a PR to remove it, just say the word.

adamralph:
Fixes #210

### DIFF
--- a/build/Program.cs
+++ b/build/Program.cs
@@ -8,7 +8,6 @@ namespace build
     class Program
     {
         private const string ArtifactsDir = "artifacts";
-        private const string Clean = "clean";
         private const string Build = "build";
         private const string RunTests = "run-tests";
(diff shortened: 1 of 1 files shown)


damianh:
Cheers lad

PR #227, opened 2018-12-06
adds logic to get access to SQL Creation scripts

Erwinvandervalk:
We use SQL Stream store in a rather corporate environment, where neither the people installing the application nor the application itself has DML rights. Only the DBA's have this access. 

So, we want to add logic to our applications that prints the latest version of the SQL Scripts that are used to create the databases, so that we can hand this off to the DBA. But, the classes that hold these scripts are internal. For most of the scripts this is fine, because they are supposed to be hidden, but for the creation scripts it would be great if we could get access to them. 

So, this adds an interface that specific implementations (MSSql, Postgres, etc.) can opt into to expose the sql creation scripts.


damianh:
I think this is a reasonable need. Am happy to have that functionality exposed on the various store implementations. I don't think it belongs on `IStreamStore` though. If you think about it this way - your application domain code will use `IStreamStore` and your infra/hosting code will know about the concrete implementations (which will have their differences).

src/SqlStreamStore/IStreamStoreSchema.cs
```diff
@@ -0,0 +1,21 @@
...
+    /// <summary>
+    /// Interface to get access to the SQL Creation scripts for each SQL based implementation of SQL Stream Store.
+    ///
+    /// This can be used when you need to run in an enterprise environment where the process doesn't have access to
+    /// do DDL statements and you need to be able to hand the DBA a script to execute. This interface allows you to
+    /// export the SQL Script, for example from the command line of your application.
+    ///
+    /// This is an optional interface that specific implementations of SqlStreamStore can choose to implement.
+    /// </summary>
+    public interface IStreamStoreSchema
```
damianh: I don't think we need this. The methods on the concrete store implementations are fine. Administrators and operators are going to be very aware of what database the scripts are intended for.

damianh:
I will add though - we are not going to support issues that arise when people export the scripts and do stuff on their own. Also... migrations will still need to be via tooling that runs a migration method and some parts are DML related and others are DDL.

src/SqlStreamStore/IStreamStoreSchema.cs
damianh: Yeah I want this interface removed; I don't see it being useful.

PR #241, opened 2019-01-04
Fix Broken Subscriptions Tests

thefringeninja:
AsyncAutoResetEvent.WaitAsync should return immediately after AsyncAutoResetEvent.Set, so these tests were not testing Dispose during the handling of a message.

Also, increasing the timeout made these tests far more likely to pass

(cherry picked from commit 88876e01ef45b1b62f3393eaa10261cd6a3ae56d)


damianh:
Cheers lad

PR #253, opened 2019-02-09
Fix For Edge Case in ExpectedVersion

thefringeninja:
Example of issue: Caller sends messages at ExpectedVersion: 0,1,2 then sends 10


damianh:
Nice.

Issue #250, opened 2019-02-01
JSONB support for Postgres


mausch:
The JSONB data type (introduced in Postgres 9.4, over 4 years ago) could be used for the messages.json_data and json_metadata columns. This enables ad-hoc queries digging into specific fields of the messages/metadata (potentially using indexes too!), adds database-level validation and is generally more efficient.

[Here's an article](https://www.compose.com/articles/faster-operations-with-the-jsonb-data-type-in-postgresql/) summarising the differences between `json` and `jsonb`. Personally I'd go with `jsonb` as it's more flexible, allowing indices on specific properties.

I applied this change to my local database after the schema was created by SqlStreamStore, and saw no issues i.e. no changes in the SqlStreamStore code were required:

```
ALTER TABLE profiles.messages 
ALTER COLUMN json_data TYPE jsonb USING json_data::jsonb,
ALTER COLUMN json_metadata TYPE jsonb USING json_metadata::jsonb;

ALTER TYPE profiles.new_stream_message
DROP ATTRIBUTE json_data,
DROP ATTRIBUTE json_metadata,
ADD ATTRIBUTE json_data jsonb,
ADD ATTRIBUTE json_metadata jsonb;
```

About json/jsonb support in clients: personally I use [dbeaver](https://dbeaver.io/), which has no issues with json/jsonb colums. Not sure about pgadmin but these json columns aren't anything new so I would expect a full support.

mickdelaney:
Any thoughts from the team on this ? 
Its a bit of a blocker for us right now.

thefringeninja:
We would definitely accept this.

thefringeninja:
Hey @mickdelaney is this working for you?

mickdelaney:
We've not integrated it yet. It's on the list for tomorrow. 
Will update here ASAP

thefringeninja:
howabout now? :smile:

mickdelaney:
Hi @thefringeninja,

Sorry for the delay. 
We're using the 1.2.0 beta 3 in a new product that is under testing right now.  Its due to go live in April. 
NOTE: 
We've gone live with 1.1 SQLStreamStore + Postgres already. 
 
So we'll upgrade that also once testing is done on the newer product. 

Regards

thefringeninja:
Fixed in #262

PR #255, opened 2019-02-12
Enforce JSON Object Literals in HTTP

thefringeninja:
This change should make #250 possible


src/SqlStreamStore.HAL/StreamMessageHALResponse.cs
```diff
@@ -0,0 +1,24 @@
...
+                message.MessageId,
+                message.CreatedUtc,
+                message.Position,
+                message.StreamId,
+                message.StreamVersion,
+                message.Type,
+                payload = FromString(payload),
+                metadata = FromString(message.JsonMetadata)
+            })
+        { }
```
damianh: I'd have done an extension method instead of inheritance

`streamMessage.ToHalResponse()`

Minor.

PR #256, opened 2019-02-13
Use try-catch instead of locks in AppendStreamExpectedVersionAny 

rikbosch:
This PR addresses a performance issue found under high load.

This lets the unique index guard the duplicates, which only happen under high concurrent load.

As posted on the slack channel.

The improvement is based on these posts

http://michaeljswart.com/2017/07/sql-server-upsert-patterns-and-antipatterns/
http://michaeljswart.com/2016/02/ugly-pragmatism-for-the-win/


damianh:
Looks promising. Thanks for PR @rikbosch

PR #258, opened 2019-02-16
suppress NU5105

adamralph:
To get rid of multiple instances of:

> warning NU5105: The package version '1.2.0-beta.3.6+build.0' uses SemVer 2.0.0 or components of SemVer 1.0.0 that are not supported on legacy clients. Change the package version to a SemVer 1.0.0 string. If the version contains a release label it must start with a letter. This message can be ignored if the package is not intended for older clients.


damianh:
Yeah that's annoying 👍

PR #259, opened 2019-02-16
switch to PackageLicenseExpression

adamralph:
gets rid of multiple instances of:

> warning NU5125: The 'licenseUrl' element will be deprecated. Consider using the 'license' element instead.


damianh:
I appear to have a timing / race condition in `When_dispose_store_then_should_dispose_all_stream_subscriptions` that is causing the build to fail. Will investigate. Gonna merge anyway, cheers.

PR #262, opened 2019-02-18
Native JSON in Postgres

mausch:
As discussed in #250

I had to make a minor change in the Postgres NewStreamMessage type as it was being fed an empty string, which is invalid JSON, which made Postgres crash. This small change is backwards compatible with the non-JSONB Postgres schema.

The alternative is fixing that in the original type, which I did in https://github.com/elevate/SQLStreamStore/commit/c112d426129e0ae8e0d0b88eda5bc8ed72b278b7 but that affected the MSSQL implementation so I had to patch things there, and I'd rather not do that. Up to you.

### DIFF
--- a/build.cmd
+++ b/build.cmd
@@ -1,5 +1,5 @@
 docker build --tag sss-build .
-docker run --rm --name sss-build ^
+docker run --rm -it --name sss-build ^
  -v /var/run/docker.sock:/var/run/docker.sock ^
  -v %cd%/artifacts:/artifacts ^
  -v %cd%/.git:/.git ^
--- a/src/SqlStreamStore.Postgres/PostgresNewStreamMessage.cs
+++ b/src/SqlStreamStore.Postgres/PostgresNewStreamMessage.cs
@@ -16,7 +16,7 @@
                 MessageId = message.MessageId,
                 Type = message.Type,
                 JsonData = message.JsonData,
-                JsonMetadata = message.JsonMetadata
+                JsonMetadata = string.IsNullOrEmpty(message.JsonMetadata) ? null : message.JsonMetadata,
             };
     }
 }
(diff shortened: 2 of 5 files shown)


src/SqlStreamStore.Postgres/PostgresNewStreamMessage.cs
```diff
@@ -16,7 +16,7 @@ public static PostgresNewStreamMessage FromNewStreamMessage(NewStreamMessage mes
                 MessageId = message.MessageId,
                 Type = message.Type,
                 JsonData = message.JsonData,
-                JsonMetadata = message.JsonMetadata
+                JsonMetadata = string.IsNullOrEmpty(message.JsonMetadata) ? null : message.JsonMetadata,
```
thefringeninja: I will need to make sure this is compatible with the HTTP API.

build.cmd
```diff
@@ -1,5 +1,5 @@
 docker build --tag sss-build .
-docker run --rm --name sss-build ^
+docker run --rm -it --name sss-build ^
```
thefringeninja: Any reason this needs to be interactive? If so, please make the same change in `build.sh` as well.

mausch: Interactive enables CTRL-C to work correctly when building in a console.
Otherwise CTRL-C cancels the batch script but leaves the docker container running in the background. Nobody wants that :)
I'll change the shell script too.

src/SqlStreamStore.Postgres/PostgresNewStreamMessage.cs
mausch: It's an internal, Postgres-specific type so it shouldn't affect any API concerns I don't think

build.cmd
mausch: Also... colours in msbuild output! :smile: 
https://travis-ci.org/SQLStreamStore/SQLStreamStore/builds/494952113#L675
vs
https://travis-ci.org/SQLStreamStore/SQLStreamStore/builds/494268682#L708

PR #266, opened 2019-03-05
[WIP] Parallelizing Tests via XUnit Collections

thefringeninja:


thefringeninja:
Tests take as long to run as before.

damianh:
They are already system resource heavy. Unlikely to see any gains unless on bigger hardware.

PR #269, opened 2019-03-19
[WIP] Catch any Inner OperationCancelledException

thefringeninja:
Some libraries (e.g. System.Data) will wrap OperationCancelledException with their own. So, unwrap them.

### DIFF
--- a/src/SqlStreamStore/ExceptionExtensions.cs
+++ b/src/SqlStreamStore/ExceptionExtensions.cs
@@ -1,0 +1,10 @@
+namespace SqlStreamStore
+{
+    using System;
+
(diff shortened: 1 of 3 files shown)


damianh:
Build still failed? :|

thefringeninja:
Yes, after 7 or 8 runs. Turns out that `OperationCancelledException` is not being wrapped, it just has the same message. I think at this point, that specific exception needs to be caught, and we throw a new  `OperationCancelledException` around it. Not ideal.

thefringeninja:
Well this is not easily solvable. This SqlException will share the same HRESULT as a lot of other things. And the message is culture sensitive.

PR #281, opened 2019-05-23
Add Optional Deletion Tracking To PG and MySql Stores

thefringeninja:
To align with MsSql V3

### DIFF
--- a/src/SqlStreamStore.MySql/MySqlStreamStoreSettings.cs
+++ b/src/SqlStreamStore.MySql/MySqlStreamStoreSettings.cs
@@ -75,5 +75,12 @@ namespace SqlStreamStore
                 _connectionFactory = value;
             }
         }
+
+        /// <summary>
+        ///     Disables stream and message deletion tracking. Will increase
+        ///     performance, however subscribers won't know if a stream or a
+        ///     message has been deleted. This can be modified at runtime.
+        /// </summary>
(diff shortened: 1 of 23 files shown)


src/SqlStreamStore.MySql/MySqlStreamStoreSettings.cs
```diff
@@ -75,5 +75,12 @@ public GetUtcNow GetUtcNow
                 _connectionFactory = value;
             }
         }
+
+        /// <summary>
+        ///     Disables stream and message deletion tracking. Will increase
+        ///     performance, however subscribers won't know if a stream or a
+        ///     message has been deleted. This can be modified at runtime.
+        /// </summary>
+        public bool DisableDeletionTracking { get; set; }
```
pgermishuys: `DisableDeletionTracking` sounds like an operation I can perform whereas `DeletionTrackingEnabled` tells  me it's a property that I am able to get/ set.

thefringeninja: Agreed, but then it would not align with what's already in there for MsSqlV3. I know that people are using the beta in production and I am reluctant to break them

pgermishuys: gotcha!

PR #283, opened 2019-06-11
Use HttpClientFactory

thefringeninja:
https://docs.microsoft.com/en-us/dotnet/standard/microservices-architecture/implement-resilient-applications/use-httpclientfactory-to-implement-resilient-http-requests

### DIFF
--- a/src/SqlStreamStore.Http/HttpClientSqlStreamStoreSettings.cs
+++ b/src/SqlStreamStore.Http/HttpClientSqlStreamStoreSettings.cs
@@ -2,20 +2,24 @@
 {
     using System;
     using System.Net.Http;
+    using Microsoft.Extensions.DependencyInjection;
     using SqlStreamStore.Infrastructure;
     using SqlStreamStore.Subscriptions;
 
     public class HttpClientSqlStreamStoreSettings
     {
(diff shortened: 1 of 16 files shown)


src/SqlStreamStore.Http/HttpClientSqlStreamStoreSettings.cs
```diff
@@ -2,21 +2,25 @@
...
+    using Microsoft.Extensions.DependencyInjection;
     using SqlStreamStore.Infrastructure;
     using SqlStreamStore.Subscriptions;
 
     public class HttpClientSqlStreamStoreSettings
     {
+        private static readonly Lazy<Func<HttpClient>> s_defaultHttpClientFactory = new Lazy<Func<HttpClient>>(() =>
+        {
+            var serviceProvider = new ServiceCollection()
+                .AddHttpClient(nameof(HttpClientSqlStreamStore)).Services
```
aarondandy: I think this could just be `.AddHttpClient()` since you don't use the typed client stuff here.

aarondandy: Oh I see, you are just naming it 🤦‍♂️ nevermind

damianh:
LGTM

PR #294, opened 2019-07-19
Fix Race Condition When Disposing MySqlDatabaseManager

thefringeninja:
ignoring condition where we find a process to kill but it's killed before we kill it


damianh:
Nice.

PR #317, opened 2019-10-03
Schema Creation Script Tool

yreynhout:
This PR provides a `dotnet tool` that allows one to generate the schema creation script for each of the SQL dialects we support. Below is an overview of the supported dialects:

- `mssqlv2` is compatible with the now obsolete `SqlStreamStore.MsSql.MsSqlStreamStore`
- `mssqlv3` is compatible with `SqlStreamStore.MsSql.MsSqlStreamStoreV3` 
- `mysql` is compatible with `SqlStreamStore.MySql.MySqlStreamStore`
- `postgres` is compatible with `SqlStreamStore.Postgres.PostgresStreamStore`

One thing I'm not happy with in this PR: The way in which the script gets generated is by instantiating the dialect matching IStreamStore implementation and calling its `GetSchemaCreationScript()` method. The problem is I need to pass a (fake) connection string to get this to work. The alternative would've been to use `InternalsVisibleTo("SqlStreamStore.SchemaCreationScriptTool")` and talk to the `Script` classes directly OR to make those script classes public. Neither of those options appealed to me at the time of authoring this, but I'm open to suggestions.

Do we want to generate the creation of the associated database schema as well? Because that does not seem to be part of what the `GetSchemaCreationScript()` method generates. It is part of the `CreateSchemaIfNotExists()` methods, though. Please note that database schema support is limited to `mssql` and `postgres` at this time.
- [x] Added a `-cs|--create-schema` command line option to cater for this (so it's opt-in).


yreynhout:
Please note that commit `6383572ed3402395489535e1f0cb4dbfacaa0cd7` will be reverted since `master` now contains a fix for the failing MySql tests.

damianh:
Rebased and merged in https://github.com/SQLStreamStore/SQLStreamStore/commit/c7a1889e9869335baadbb6c88357c89fda59fd56

PR #329, opened 2019-10-29
Run tests projects in serial instead of parallel.

damianh:
Attempt to resolve flake test issues because, I think, the system load using parallel tests is causing timeouts and resulting in CI failures.


damianh:
Looks like that worked. Going to re-run the tests a couple of times. 

The cost is that it runs a few minutes slower.

PR #330, opened 2019-10-30
Database Fixture pooling

damianh:
Instead of droping and recreating the database for every test, we reuse previously created database and reset them (delete all rows, reset increment seeds) instead.

## MsSQL
- Pools are based on schema name. 
- If two tests uses default `[dbo]` and run sequentially, the database will be reused.
- If two tests uses default `[dbo]` and run concurrently, a new database will be deployed.
- Only works for tests run in the same process.
- Some tests assert schema related things, these are using unique schema to prevent pooled usage.
- It is expected there will be 10s of databases in the pool depending on degree of test parallism.
- As state is now (somewhat) shared, extra care is needed. Hopefully the performance benefits will be worth it.
- Individual test runs will obviously not benefit.

Other dbs to follow...


damianh:
Before,6m: 
image

After, <1m (but with a couple of failing tests):
image

Looks promising!

damianh:
5 mins to run the tests on Travis

image

Probably faster if no failures....

damianh:
There is tests that is failing more consistently now this runs faster. Merging and addressing that is a separate PR.

PR #365, opened 2020-02-25
Fix typos in README.md

wooderz:
The nuget links were mistyped as muget - They don't go anywhere so fixing to nuget


damianh:
Looks like muget died 🤷

damianh:
Thx for PR

PR #369, opened 2020-03-03
canonical off switch stream resource

vmachacek:
related to https://github.com/SQLStreamStore/SQLStreamStore/issues/368

- adding support for **UseCanonicalUrls** to **StreamResource**

### DIFF
--- a/src/SqlStreamStore.HAL/Streams/StreamResource.cs
+++ b/src/SqlStreamStore.HAL/Streams/StreamResource.cs
@@ -11,14 +11,16 @@ namespace SqlStreamStore.HAL.Streams
     {
         private readonly IStreamStore _streamStore;
         private readonly string _relativePathToRoot;
+        private readonly bool _useCanonicalUrls;
         public SchemaSet Schema { get; }
 
-        public StreamResource(IStreamStore streamStore)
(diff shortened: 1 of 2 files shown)


src/SqlStreamStore.HAL/Streams/StreamResource.cs
```diff
@@ -11,14 +11,16 @@ internal class StreamResource : IResource
     {
         private readonly IStreamStore _streamStore;
         private readonly string _relativePathToRoot;
+        private readonly bool _useCanonicalUrls;
         public SchemaSet Schema { get; }
 
-        public StreamResource(IStreamStore streamStore)
+        public StreamResource(IStreamStore streamStore, bool useCanonicalUrls)
```
damianh: I merged this a bit quick. Should this have a default value?

thefringeninja: no, cause it's only newed up in one place.

vmachacek: I was looking at `AllStreamResource` which have same signature, but doesn't have param as optional. The constructor is used only once in a codebase.

Issue #377, opened 2020-03-26
Trouble upgrading from 1.1.3 to 1.2.0-beta.8


yreynhout:
You have an app A that depends on component B and component B depends on `>= 1.1.3` of SqlStreamStore. Now, app A itself depends on `1.2.0-beta.8` of SqlStreamStore. Build and publish work as expected, that is the dependency becomes `1.2.0-beta.8`.`1.1.3` has as its `AssemblyVersion` ... `1.1.3.0`.

`MinVer` was introduced as part of `1.2.0` development. Due to how `MinVer` hands out the `AssemblyVersion` (as per its documentation https://github.com/adamralph/minver#version-numbers) it has set it to `1.0.0.0`.

This causes a `FileLoadException` at runtime when app A is started and invokes types in component B that use SqlStreamStore, stating that it's looking for `1.1.3.0` which it can't find, because, well, there's only `1.0.0.0`.

```
Could not load file or assembly 'SqlStreamStore, Version=1.1.3.0, Culture=neutral, PublicKeyToken=null'. The located assembly's manifest definition does not match the assembly reference. (0x80131040))\n ---> System.IO.FileLoadException: Could not load file or assembly 'SqlStreamStore, Version=1.1.3.0, Culture=neutral, PublicKeyToken=null'. The located assembly's manifest definition does not match the assembly reference. (0x80131040)\nFile name: 'SqlStreamStore, Version=1.1.3.0, Culture=neutral, PublicKeyToken=null
```

@adamralph @damianh

yreynhout:
The resolution _might_ be to take over the assembly version assignment from `MinVer` and set it to `1.2.0.0` once it is released and keeping it for the entire lifecycle of `1.2+`

adamralph:
It sounds like you should keep the old versioning scheme for `AssemblyVersion` for the 1.x range:

```xml
<Target Name="MyTarget" AfterTargets="MinVer" Condition="'$(MinVerMajor)' == '1'">
  <PropertyGroup>
    <AssemblyVersion>$(MinVerMajor).$(MinVerMinor).$(MinVerPatch).0</AssemblyVersion>
  </PropertyGroup>
</Target>
```

yreynhout:
Addressed by #378

PR #378, opened 2020-03-27
Fix assemblyversion for version 1 releases (minver tweak)

yreynhout:
This PR addresses the issue raised in #377 using the solution proposed by @adamralph

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -20,4 +20,9 @@
     <PackageReference Include="MinVer" Version="2.0.0" PrivateAssets="All" />
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0" PrivateAssets="All"/>
   </ItemGroup>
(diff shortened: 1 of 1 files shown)


yreynhout:
Proof from a local run (extracted dll from a nupkg)

image

yreynhout:
Reflective question: how will this affect components already depending on 1.2.0-beta.x? The requested assembly version will be `1.0.0.0` whereas this change will turn that assembly version into `1.2.0.0`. Thoughts, @adamralph @damianh @thefringeninja ?

adamralph:
@yreynhout doesn't the implicit assembly redirecting take care of that now? I.e if it's a higher version it will just take it?

yreynhout:
I'll have a look at this using a local feed to see how it behaves.

yreynhout:
Just tried this with the local build output, i.e. have an app that depends on 1.2.0-beta.8.71 and a lib that depends on 1.1.3 and have the app depend on the lib and use the lib's type which exposes a SqlStreamStore member ... it seems to be working. Going to merge for now, but we probably want some CI feed beta testers before releasing.

PR #383, opened 2020-04-11
Fix InMemoryStreamstore reading a stream backwards with gaps

rvdkooy:
This PR will fix an issue in the InMemoryStreamStore `ReadStreamBackwardsInternal` where it cannot find all messages anymore when there a gaps in the stream (caused by delete operations).

questions:
- Could this new approach cause any performance issues?
- Is this only a InMemoryStreamStore issue? Can this issue also occur in the other providers? (I haven't checked this)
- I'm not sure if the spec that I wrote is in the correct location (I didn't find any explicit inmemory specs yet)

### DIFF
--- a/src/SqlStreamStore/InMemory/InMemoryStreamStore.cs
+++ b/src/SqlStreamStore/InMemory/InMemoryStreamStore.cs
@@ -601,41 +601,37 @@ namespace SqlStreamStore
                     return Task.FromResult(notFound);
                 }
 
-                var messages = new List<StreamMessage>();
-                var i = fromVersionInclusive == StreamVersion.End ? stream.Messages.Count - 1 : fromVersionInclusive;
-                while (i < stream.Messages.Count && i >= 0 && count > 0)
+                var fromVersion = fromVersionInclusive == StreamVersion.End ? stream.Messages.Max(m => m.StreamVersion) : fromVersionInclusive;
+
+                var messages = stream.Messages.Where(m => m.StreamVersion <= fromVersion).Reverse().Take(count).Select(m =>
                 {
-                    var inMemorymessage = stream.Messages[i];
                     StreamMessage message;
                     if (prefetch)
                     {
                         message = new StreamMessage(
                             streamId,
-                            inMemorymessage.MessageId,
-                            inMemorymessage.StreamVersion,
-                            inMemorymessage.Position,
-                            inMemorymessage.Created,
-                            inMemorymessage.Type,
-                            inMemorymessage.JsonMetadata,
-                            inMemorymessage.JsonData);
+                            m.MessageId,
+                            m.StreamVersion,
+                            m.Position,
+                            m.Created,
+                            m.Type,
+                            m.JsonMetadata,
+                            m.JsonData);
                     }
                     else
                     {
                         message = new StreamMessage(
                             streamId,
-                            inMemorymessage.MessageId,
-                            inMemorymessage.StreamVersion,
-                            inMemorymessage.Position,
-                            inMemorymessage.Created,
-                            inMemorymessage.Type,
-                            inMemorymessage.JsonMetadata,
-                            ct =>  Task.Run(() => ReadMessageData(streamId, inMemorymessage.MessageId), ct));
+                            m.MessageId,
+                            m.StreamVersion,
+                            m.Position,
+                            m.Created,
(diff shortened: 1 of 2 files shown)


yreynhout:
The general idea with acceptance tests is that they are applied across all providers (unless there's a compelling incentive not to). What exactly was the problem with reading a stream backwards before this fix (just curious)? That'll help me assess where the test needs to go.

rvdkooy:
When reading the stream backwards: an indexer is used to lookup the message(s).

The indexer was based on:
`var i = fromVersionInclusive == StreamVersion.End ? stream.Messages.Count - 1 : fromVersionInclusive;` 

The lookup with this indexer does not work anymore if you remove messages from the stream

example
``` 
messages[0] = Message { StreamVersion: 0 }
messages[1] = Message { StreamVersion: 1 }
// message with StreamVersion 2 was deleted
messages[2] = Message { StreamVersion: 3 }
messages[3] = Message { StreamVersion: 4 }
```
to lookup the message with streamVersion: 3 it would do:

```
var inMemorymessage = stream.Messages[i]; // ==> i = StreamVersion 3
```
this is wrong because message with StreamVersion 3 is actually at position 2 now, insteadof 3

I hope explaination clears it up

yreynhout:
Sorry to keep this lingering - will review later today.

tests/SqlStreamStore.AcceptanceTests/AcceptanceTests.ReadStream.cs
```diff
@@ -372,6 +372,24 @@ public async Task When_read_stream_backwards_with_url_encodable_stream(string st
...
         }
 
+        [Fact, Trait("Category", "ReadStream")]
+        public async Task When_reading_a_stream_backwards_that_has_gaps_and_providing_the_fromVersion_explicitly()
+        {
+            var streamId = new StreamId("streamid");
+            var messages = CreateNewStreamMessages(1, 2, 3);
+            var firstMessageId = messages[0].MessageId;
+            var secondMessageId = messages[1].MessageId;
+            var ThirdMessageId = messages[2].MessageId;
```
yreynhout, line 382: Minor, casing.

src/SqlStreamStore/InMemory/InMemoryStreamStore.cs
```diff
@@ -600,42 +600,40 @@ protected override Task<ReadStreamPage> ReadStreamForwardsInternal(string stream
...
+                            m.JsonMetadata,
+                            ct =>  Task.Run(() => ReadMessageData(streamId, m.MessageId), ct));
                     }
-                    messages.Add(message);
-
-                    i--;
-                    count--;
-                }
+                    return message;
+                }).ToList();
```
yreynhout, line 636: I'd prefer not to lean on LINQ so much to be honest - not sure how it affects peformance.

src/SqlStreamStore/InMemory/InMemoryStreamStore.cs
```diff
@@ -600,42 +600,40 @@ protected override Task<ReadStreamPage> ReadStreamForwardsInternal(string stream
                         s_readNextNotFound);
                     return Task.FromResult(notFound);
                 }
+                
+                var fromVersion = (fromVersionInclusive == StreamVersion.End) ?
```
yreynhout, line 604: Think I'm gonna need a drink after reading this one ...

src/SqlStreamStore/InMemory/InMemoryStreamStore.cs
rvdkooy: I have another idea how to fix this
This does not involve LINQ and should also be much faster

will come back on this

thanks

rvdkooy: My next approach replaces the `List<InMemoryMessage>` with a `SortedList<StreamVersion, InMemoryMessage>` as the underlying datastructure. With this we can still use the indexer approach as before (so no LINQ 👍 ) plus the ability to find the `actual` index by calling: `IndexOfKey` on the SortedList

Also the `Messages` property on the `InMemoryStream` is removed, because I couldn't make it readonly anymore when being exposed. For that I added explicit Methods and Props like: `GetMessage`, `NumberOfMessages`, `GetLastMessage`, `GetIndexOfStreamVersion`

PR #385, opened 2020-04-20
Sqlite Implementation

dealproc:
Implementation of https://github.com/SQLStreamStore/SQLStreamStore/issues/28 .


damianh:
Epic PR @dealproc !!

There are a few things I'd like to tweak but I've merged as-is to get into `master` and part of CI.

Cheers mate 😄

claudiuchis:
Hi folks, is the Sqlite implementation going to be packaged into a nuget package, similar to MsSql and the other variants? Or, is there another way to use this? Many thanks for this great work! (I'm using the InMemory implementation and looking to use the MsSql for Azure Sql)

PR #387, opened 2020-04-27
Introduce command timeout as a mssql setting and apply to sqlcommands

yreynhout:
This PR introduces `CommandTimeout` as a setting of the MsSql Stream Store implementation, allowing callers to control the command timeout of a `SqlCommand`.


CumpsD:
Awesome, thanks!

PR #389, opened 2020-04-28
Fix the build

damianh:
A bunch of work to get the build experience pristine again.

- Clean up github workfows: remove obsolete actions & non-shallow clone.
- DropAll  scripts will test for existence of tables.
- Dropping tables and recreating tables instead of clearing down exisiting data. (Was having inssues with identity colums reseeding correctly). No performance loss as it's creating the db in the first place that is expensive.
- MySql using BeginTransactionAsync where possible.
- Replace Docker.Net with FluentDocker and cleaned up container start and availability checks.
- MySql has a bit of a crude availability check that set the max_connections. Can be optimised later, works for now.
- Fix paths when starting the build container so artifacts are written back to the host.
- Build docker file is using damianh/dotnet-core-lts-sdks:1 because it needs netcoreapp3.1 and 2.1 installed.
- Some typos here and there.

Note:
- Skipped some tests that appear to be experiencing a race condition in CI. To be followed up on next.


tests/SqlStreamStore.MsSql.V3.Tests/MsSqlStreamStoreV3Fixture.cs
```diff
@@ -69,32 +69,32 @@ public async Task Prepare()
             { 
                 using (var connection = new SqlConnection(_settings.ConnectionString))
                 {
-                    connection.Open();
+                    await connection.OpenAsync();
```
yreynhout, line 72: Do we need ConfigureAwait(false) sprinkled here?

damianh: I know one needs such in shipped libraries. I don't think it's necessary in test projects.

damianh:
Definitely made progress on this. Going to squash it when done.

damianh:
Nearly there...

This tests fails fairly consistently. Smells of race condition:

damianh:
Ok the HAL tests weren't running because SDK wasn't in the build continer. Using a different image now.

Also, local green build!

image

PR #390, opened 2020-05-01
Fix for default fetch depth of actions/checkout@v2 

rvdkooy:
actions/checkout@v2 uses a fetch depth of 1 for performance optimizations. This will not work with minver see:
https://github.com/adamralph/minver#why-is-the-default-version-sometimes-used-in-github-actions-and-travis-ci-when-a-version-tag-exists-in-the-history

This PR sets to fetch depth to 100. This should be enough to find the latest tag in the history.

R

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -16,6 +16,8 @@ jobs:
     steps:
     - name: Checkout
       uses: actions/checkout@v2
+      with:
+        fetch-depth: '100'
     - name: Generate build number
       uses: einaregilsson/build-number@v2
       with:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -17,7 +17,7 @@ jobs:
     - name: Checkout
       uses: actions/checkout@v2
       with:
-        fetch-depth: '100'
+        fetch-depth: '0'
     - name: Generate build number
       uses: einaregilsson/build-number@v2
       with:
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -11,7 +11,7 @@ jobs:
     - name: Checkout
       uses: actions/checkout@v2
       with:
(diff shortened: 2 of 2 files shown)


yreynhout:
Good catch, though as per the documentation, I might be inclined to go for `fetch-depth: 0` to see how it affects build performance (this is not a high transaction volume repository).

rvdkooy:
yes, that's probably the safest choice. I updated it to 0

R

thefringeninja:
I think that `actions/checkout@v2` is part of the problem here. `actions/checkout@af513c7a016048ae468971c52ed77d9562c7c819` seems to work without setting fetch depth.

yreynhout:
We switched from `v1` to `v2` to overcome the problem of being able to re-run your GHA workflows. `v2` seems to have brought a change that is incompatible with `minver`, which is why I was under the assumption that settings `fetch-depth` to `0` would make `minver` happy again.

thefringeninja:
It doesn't unfortunately as I've discovered. However this seems to work:
```
    - name: Checkout
      uses: actions/checkout@v2
    - shell: bash
      run: |
        git fetch --prune --unshallow
    - name: MinVer
      shell: bash
      run: |
        dotnet tool restore
        echo $(dotnet tool run minver)
```

rvdkooy:
After this PR I now see this:

`/usr/bin/git -c protocol.version=2 fetch --no-tags --prune --progress --no-recurse-submodules origin +b318c1892754a81bc1abe9cb0eda0cc706ef36ff:refs/remotes/origin/master`

where it previously did:

`/usr/bin/git -c protocol.version=2 fetch --no-tags --prune --progress --no-recurse-submodules --depth=1 origin +fc7d40dbc13b3e89c7b2d04d73df529f73a79b29:refs/remotes/pull/391/merge`

notice the: `--depth=1` is gone now, what do you get?

R

PR #391, opened 2020-05-04
ReadStreamHeadPosition/Version

yreynhout:
This PR introduces two new methods on `IReadonlyStreamStore` (non-breaking change):

- `ReadStreamHeadPosition`: reads the head position of a stream, i.e. the position of its last message
- `ReadStreamHeadVersion`: reads the head version of a stream, i.e. the version of its last message

The test suite was extended with additional acceptance tests to cover the behavior of these new operations. This is the basis for a follow-up PR (in that it will follow a similar pattern) which will introduce support for reading the stream head message as well as the all stream head message.

### DIFF
--- a/src/SqlStreamStore.Http/HttpClientSqlStreamStore.cs
+++ b/src/SqlStreamStore.Http/HttpClientSqlStreamStore.cs
@@ -70,6 +70,44 @@
             return headPosition;
         }
 
+        public async Task<long> ReadStreamHeadPosition(StreamId streamId, CancellationToken cancellationToken = default)
+        {
+            GuardAgainstDisposed();
+
+            var client = CreateClient();
+            var response = await client.Client.HeadAsync(LinkFormatter.ReadStreamBackwards(streamId, StreamVersion.End, 1, false), cancellationToken);
+
+            response.EnsureSuccessStatusCode();
+
+            var resource = client.Current.First();
+
+            if(client.StatusCode == HttpStatusCode.NotFound)
+            {
+                return -1L;
+            }
+
+            return resource.Data<HalReadPage>().LastStreamPosition;
+        }
+
+        public async Task<int> ReadStreamHeadVersion(StreamId streamId, CancellationToken cancellationToken = default)
+        {
+            GuardAgainstDisposed();
+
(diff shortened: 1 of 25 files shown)


src/SqlStreamStore.Http/HttpClientSqlStreamStore.cs
```diff
@@ -70,6 +70,20 @@ public async Task<long> ReadHeadPosition(CancellationToken cancellationToken = d
             return headPosition;
         }
 
+        public async Task<long> ReadStreamHeadPosition(StreamId streamId, CancellationToken cancellationToken = default)
+        {
+            GuardAgainstDisposed();
+
+            return (await ReadStreamBackwards(streamId, StreamVersion.End, 1, false, cancellationToken)).LastStreamPosition;
```
yreynhout, line 77: If this could be solved in a more optimized way for the http provider, please show me how.

src/SqlStreamStore.Http/HttpClientSqlStreamStore.cs
```diff
@@ -70,6 +70,20 @@ public async Task<long> ReadHeadPosition(CancellationToken cancellationToken = d
...
+        }
+
+        public async Task<int> ReadStreamHeadVersion(StreamId streamId, CancellationToken cancellationToken = default)
+        {
+            GuardAgainstDisposed();
+
+            return (await ReadStreamBackwards(streamId, StreamVersion.End, 1, false, cancellationToken)).LastStreamVersion;
```
yreynhout, line 84: If this could be solved in a more optimized way for the http provider, please show me how.

src/SqlStreamStore.MsSql/MsSqlStreamStore.cs
```diff
@@ -280,13 +281,63 @@ protected override async Task<long> ReadHeadPositionInternal(CancellationToken c
 
                     if(result == DBNull.Value)
```
yreynhout, line 282: I've left this in - but the test for the new operations seem to suggest this path will never be hit (since result is either `null` or a value, not `DBNull.Value`). Similar remark in other relational store implementations.

PR #392, opened 2020-05-07
Fix Issue w/ Fetching Tags Incorrectly

thefringeninja:


rvdkooy:
@thefringeninja, I think they fixed the real issue here:
https://github.com/actions/checkout/commit/e52d022eb52c224e5f2201beb687a66849e3b200

PR #398, opened 2020-05-15
PR to analyze deadlock when appending rows in MySql. 

Erwinvandervalk:
There appears to be a deadlock when inserting rows in mysql.

This PR adds the option to run the load tests against mysql, then also implements a specific load test that exposes this.


damianh:
Merged. Will address the issue in a follow-up PR.

PR #409, opened 2020-05-29
Parallelize Builds with Matrix

thefringeninja:


damianh:
While it doesn't make it any faster, I like surfacing the checks like this.

Issue #406, opened 2020-05-22
GetSchemaCreationScript() should include schema (namespace) creation 


rofr:
MsSql and Postgres have the concept of schemas. Schemas are isolated namespaces within a database. SqlLite and MySql have no such concept. (MySql does have a SCHEMA keyword but it is a synonym for DATABASE.)

SqlStreamStore supports schemas for both mssql and postgres by way of the properties `MsSqlStreamStoreSettings.Schema`, `MsSqlStreamStoreV3Settings.Schema` and `PostgresStreamStoreSettings.Schema`.

The schema is created automatically when calling `CreateSchemaIfNotExists()` but not included in the script returned by `GetSchemaCreationScript()` This means the schema must be created separately before executing the generated script or it will fail.

I suggest the schema creation logic be included in the generated script so it will have the same behavior as calling `CreateSchemaIfNotExists()` This should apply to mssql and postgres as it is not applicable to the other providers.

PR #410, opened 2020-05-30
Include CREATE SCHEMA in scripts

rofr:
Postgres and Mssql providers support custom schemas. The schemas are created when calling `CreateSchemaIfNotExists()` with hard coded SQL, while the rest of the objects are created based on embedded resource SQL scripts. The scripts can be extracted with `GetSchemaCreationScript()` but will not include the CREATE SCHEMA logic. This PR remedies that behavior. #406


yreynhout:
This PR seems to duplicate parts of  #317 IMO.

rofr:
> This PR seems to duplicate parts of #317 IMO.

Yes, #317 is affected. This PR targets the root cause and eliminates the need for a workaround in the schema creation tool. I suggest that the --create-schema option in the tool be removed.

yreynhout:
It's not really a workaround, TBH. It's opt-in - I don't see that option here.

rofr:
> It's not really a workaround, TBH. It's opt-in - I don't see that option here.

But the option is not necessary if the script contains an idempotent CREATE SCHEMA statement. Qith this change it will become obsolete and provide no extra value. If you want to include a CREATE SCHEMA it is there and if you want to create the schema separately, the script will work just as well.

PR #427, opened 2020-07-06
remove build number from version

adamralph:
I noticed this:

image

The GitHub Actions workflow is not setting `BUILD_NUMBER` so it's always the default value of `0`. Thing is, GitHub Actions doesn't have build numbers. For that reason, I've stopped putting the build number into the versions of all my products. It never really added much value any way.


damianh:
> Thing is, GitHub Actions doesn't have build numbers

Just to say, there is a `GITHUB_RUN_NUMBER` env var. https://docs.github.com/en/actions/configuring-and-managing-workflows/using-environment-variables

adamralph:
👍 @damianh I think that must be new. When I switched to GHA I think all they had was `GITHUB_RUN_ID` which is a random number.

PR #433, opened 2020-07-09
Mysql ReadAllForward improvements

### LINKED ISSUE #432: Performance issue when ReadingAllForwards on MySql
Lifted the current ReadAll sproc to workbench on a sample data set and it appears to be doing a Full Table Scan and results in 5-10 seconds:

image

Simplified the query removing the `CASE` statement results in an Index Range Scan and results in 0s:

image

The MSSQL implementation uses independent, non sproc, statements for various ReadAll permutaions, thus changes are non-breaking. However changes to MYSQL spocs will require a schema migration. On the other hand, if I just use standard statements and ingore the sproc, it won't require a schema change.

@thefringeninja Do you see an particular disadvantages to useing standard statements instead of sprocs?

damianh:
Fixes #432 

- Created seperate statement sql files for `ReadAllForward` and `ReadAllForwardWithData` that replaces the one stored proceedure.
- The ReadAllForward sql now use `STRAIGH_JOIN` as it appears the MySql optimiser was doing an incorrect join resulting in the slow read perf.
- No changes to schema, but there will be an unused sproc. This will be removed in a future schema version with migration
- Similar work will need to be applied to for other reads; they will come in subsequent PRs.

### DIFF
--- a/src/SqlStreamStore.MsSql/MsSqlStreamStoreV3.ReadAll.cs
+++ b/src/SqlStreamStore.MsSql/MsSqlStreamStoreV3.ReadAll.cs
@@ -36,7 +36,6 @@ namespace SqlStreamStore
                         .ExecuteReaderAsync(CommandBehavior.SequentialAccess, cancellationToken)
                         .NotOnCapturedContext();
 
-                    var messages = new List<(StreamMessage, int?)>();
                     if (!reader.HasRows)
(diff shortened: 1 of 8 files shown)


damianh:
Will be tagging this `v.1.2.1-alpha.1` for upstream validation.

PR #435, opened 2020-07-11
Added GrantRead & GrantAppend  operations to MsSqlStreamStoreV3

ylorph:
This adds 2 methods `GrantRead` & `GrantAppend` to the MsSqlStreamStoreV3 connection to allow to grant necessary rights for a sql user to read or write to streams.

They add user to  the `db_datareader` & `db_datawriter` fixed db group + grant execute to `NewStreamMessages` table type .
This gives broad access to the DB where the streams are created, but  should cover most needs 

The grant execute  on `NewStreamMessages`  does create a sql through string replacement, it's not possible to pass the full scheme+table type  & user through parameters afaik  
On way to avoid that we would be to use Microsoft.SqlServer.SqlManagementObjects package, but it means to always take that dependency for just 1 of the operations

Other approach are possible for more granularity on the security side, 
e.g create new database role with access only to the stream & message releated tables & the talbe type, and  add users to those', 

but that is a choice that is best left to the consumer of the library .

### DIFF
--- a/tests/SqlStreamStore.MsSql.V3.Tests/MsSqlStreamStoreV3AcceptanceTests.cs
+++ b/tests/SqlStreamStore.MsSql.V3.Tests/MsSqlStreamStoreV3AcceptanceTests.cs
@@ -3,6 +3,7 @@
     using System;
     using System.Diagnostics;
     using System.Threading.Tasks;
+    using Microsoft.Data.SqlClient;
     using Shouldly;
     using SqlStreamStore.Streams;
     using Xunit;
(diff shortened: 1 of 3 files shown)


tests/SqlStreamStore.MsSql.V3.Tests/MsSqlStreamStoreV3AcceptanceTests.cs
```diff
@@ -103,5 +104,74 @@ public async Task Time_taken_to_append_and_read_large_message_without_prefetch()
...
+                await fixture.Store.AppendToStream(streamId, ExpectedVersion.Any, newStreamMessage);
+
+                var store = fixture.StoreFor("can_not_read");
+                var exception = await Record.ExceptionAsync(() => store.ReadStreamForwards(streamId, StreamVersion.Start, 1));
+                exception.ShouldBeOfType<SqlException>()
+                    .Message.ShouldContain("The SELECT permission was denied");
+            }
+        }
+        [Fact]
+        public async Task Can_Not_Read_When_Rights()
```
bartelink, line 125: remove _Not ;)

PR #468, opened 2020-10-04
Replace NotOnCapturedContext with ConfigureAwait(false) (fixes #436)

### LINKED ISSUE #436: Replace .NotOnCapturedContext(); with standard .ConfigureAwait(false);
While it better describes what is actually happening, it's not standard so adds friction to codebase for new eyes.

huysentruitw:


damianh:
Yeah idomatic ftw. What was I thinking :)

PR #480, opened 2020-11-25
Add a failing concurrency test

roetlich:
Add a test to AcceptanceTests.AppendStream that fails for postgres but passes for all other DBs. This is to reproduce a bug with concurrent appends in postgres, see: https://github.com/SQLStreamStore/SQLStreamStore/issues/478


thefringeninja:
@roetlich can you rebase your PR please?

stoft:
@thefringeninja rebased.

Mousaka:
Me and @EddieChoCho tried to find a solution for this but has not succeeded yet. However, we found out that it's this exception that is thrown here:
https://github.com/SQLStreamStore/SQLStreamStore/blob/master/src/SqlStreamStore.Postgres/PgSqlScripts/EnforceIdempotentAppend.sql#L34

thefringeninja:
Sorry for the late reply. Is this something that fails occasionally?

roetlich:
When I made this PR, the test was failing consistently for me locally. AFAIR it also failed in the CI here in GitHub, but it was shown as a passing check by some mistake. But it's been a few month, but maybe with current code that's not the case anymore and I might be misremembering something.

thefringeninja:
Part of the problem is that I can't re-run the checks associated with the PR. I guess it's because you are not a committer? I'll take a closer look at this soon

Mousaka:
It was consistently failing for me locally as well when I was running it

roetlich:
I rebased this on top of your master, so now the CI was retriggered. Now it even shows which test is failing! 😁

roetlich:
@thefringeninja is there any update on this? Can I help you with testing this in any way?

